PR #191, opened 2026-04-14
fix(pg-delta): supabase add support for pgmq extension

avallete:
## Summary
Fixes declarative schema apply failing for projects using the **`pgmq`** extension when the Supabase integration exports extension DDL that includes `WITH SCHEMA pgmq`, while **`CREATE SCHEMA pgmq` is not part of the exported set** (system schema filtering). Also formalizes **`SerializeOptions`** so integration serialize rules and `Change.serialize()` share one typed contract, documents those types in TypeDoc, and adds a regression test for the export → apply pipeline.
## Problem
PostgreSQL requires the target schema to **already exist** before `CREATE EXTENSION … WITH SCHEMA <schema>`. On a **fresh** database (e.g. shadow DB / clean apply), the `pgmq` schema typically **does not exist yet**. The `pgmq` extension creates that schema as part of installation, but the server validates `WITH SCHEMA` **before** running the extension scripts—so `CREATE EXTENSION pgmq WITH SCHEMA pgmq` fails with `schema "pgmq" does not exist` unless something else has pre-created `pgmq`.
Separately, wiring serialize options through the integration DSL exposed a typing gap: the `Change` union did not consistently accept `serialize(options)` until concrete change types aligned on a shared options type.
## Solution
- **Supabase integration:** Serialize rule for `pgmq` extension creates so emitted SQL omits `WITH SCHEMA` (plain `CREATE EXTENSION pgmq;`), avoiding the invalid ordering on a fresh database when the schema create is not exported.
- **Core typing:** Central `SerializeOptions` in `serialize.types.ts`, used by `compileSerializeDSL`, `BaseChange.serialize`, and concrete serializers; `CreateExtension` honors `skipSchema` where applicable.
- **Docs:** Re-export serializer option types from the TypeDoc entrypoint and add TSDoc so they appear in the generated API reference.
- **Tests:** Integration regression covering declarative export + apply roundtrip for a Supabase-shaped `pgmq` setup; unit coverage for DSL / extension serialization behavior.
- **Changeset:** Patch bump for `@supabase/pg-delta`.


avallete (commented):
Most of the changes are about adding a `_options: SerializeOptions` into every serialize change implementation, so downstream dsl filter can have some typing correctly infered.

Real changes are in:

`packages/pg-delta/src/core/integrations/supabase.ts`
`packages/pg-delta/tests/integration/pgmq-declarative-roundtrip.test.ts`

jgoux (approved):

PR #192, opened 2026-04-14
fix(pg-delta): drop out of order operations

### LINKED ISSUE #193: Drop out of order with FK constraints
> I’ve been running into several dependency-related issues in pg-delta plan (export sql), especially around DROP ordering and change semantics, and I’m trying to understand whether these are expected to be addressed as part of the SortingCatalog refactor (#115), or if they should be handled separately.
> 
> ### 1. Missing table-level dependencies for foreign keys (DROP order)
> 
> When dropping two tables with an FK relationship:
> 
> ```sql
> CREATE TABLE challenge_levels (
>   id BIGSERIAL PRIMARY KEY
> );
> 
> CREATE TABLE user_challenge_progress (
>   id BIGSERIAL PRIMARY KEY,
>   level_id BIGINT REFERENCES challenge_levels(id)
> );
> ```
> Dropping both results in incorrect ordering:
> ```sql
> DROP TABLE challenge_levels;
> DROP TABLE user_challenge_progress;
> ```
> Which fails with:
> ```
> ERROR: cannot drop table challenge_levels because other objects depend on it
> DETAIL: constraint user_challenge_progress_level_fkey depends on table challenge_levels
> ```
> This seems to be because FK dependencies exist at the constraint level,
> but are not lifted to table-level dependencies for sorting.
> 
> ### 2. Operation-level dependencies (e.g. ALTER PUBLICATION ... DROP TABLE)
> Cases like:
> ```sql
> CREATE TABLE challenge_levels (
>   id BIGSERIAL PRIMARY KEY
> );
> 
> CREATE PUBLICATION pub FOR TABLE challenge_levels;
> ```
> Current generated order (incorrect):
> ```sql
> DROP TABLE challenge_levels;
> ALTER PUBLICATION pub DROP TABLE challenge_levels;
> ```
> This fails because the table must still exist when altering the publication:
> ```
> ERROR: relation "challenge_levels" does not exist
> ```
> Expected correct order:
> ```
> ALTER PUBLICATION pub DROP TABLE challenge_levels;
> DROP TABLE challenge_levels;
> ```
> 
> ### Question
> 
> I’ve run into several DROP ordering issues , and I’m trying to understand how they relate to this refactor.
> 
> Are these kinds of issues expected to be resolved as part of the SortingCatalog work?
> 
> Or are they considered out of scope, and should be handled separately (e.g. via additional issues in pg-delta)?
> 
> If they are not fully covered by this refactor, I’m happy to open separate issues for each case — just wanted to confirm the intended scope first. 

 _Originally posted by @Zohix1 in [#115](https://github.com/supabase/pg-toolbelt/issues/115#issuecomment-4235308639)_

avallete:
I dug into the reported DROP-ordering regressions and reproduced both cases with focused integration tests before changing sorter behavior.

## Context

There were two concrete failures in generated plan SQL:

1. FK-related table drops were ordered as:
   - `DROP TABLE challenge_levels`
   - `DROP TABLE user_challenge_progress`

   That fails because the FK on `user_challenge_progress` still depends on `challenge_levels`.

2. Publication membership changes were ordered as:
   - `DROP TABLE challenge_levels`
   - `ALTER PUBLICATION pub DROP TABLE challenge_levels`

   That fails because the table must still exist when removing it from the publication.

## Root Cause

The interesting part is that the catalog dependency extraction was already capturing the important relationships:
- FK dependencies exist at the constraint level
- publication-to-table dependencies already exist in the dependency data

The actual gap was in the change metadata that feeds the sorter:

- `DropTable` only exposed table and column stable IDs, so FK edges that only existed at the constraint level were not influencing whole-table drop ordering.
- `AlterPublicationDropTables` exposed `requires`, but not `drops`, so it was treated like a non-destructive `ALTER` and sorted after the real table drop.

## Fix

This patch fixes the ordering by making the destructive changes advertise the right stable IDs to the sorter:

- `DropTable` now includes table constraint stable IDs in its drop-phase metadata, which lets FK constraint dependencies affect whole-table drop ordering.
- `AlterPublicationDropTables` now exposes dropped table stable IDs, so it participates in the destructive/drop phase and runs before `DROP TABLE`.

I also added focused integration coverage for both reported scenarios, plus a small unit assertion around the publication change metadata.

## Verification

Verified with:
- `PGDELTA_TEST_POSTGRES_VERSIONS=17 bun run test tests/integration/fk-constraint-ordering.test.ts`
- `PGDELTA_TEST_POSTGRES_VERSIONS=17 bun run test tests/integration/publication-operations.test.ts`
- `PGDELTA_TEST_POSTGRES_VERSIONS=17 bun run test tests/integration/trigger-operations.test.ts`
- `bun run test src/core/objects/publication/changes/publication.alter.test.ts`
- `bun run test src/core/objects/table/changes/table.drop.test.ts`

So this ends up being less about missing raw dependency extraction and more about making sure destructive change objects surface enough information for the sorter to use the dependency graph correctly.

Fixes #193

### DIFF
--- a/.changeset/fix-drop-ordering.md
+++ b/.changeset/fix-drop-ordering.md
@@ -1,0 +1,5 @@
+---
+"@supabase/pg-delta": patch
+---
+
+fix(sort): order FK-related table drops and publication table removals before dependent destructive operations
--- a/packages/pg-delta/src/core/objects/publication/changes/publication.alter.test.ts
+++ b/packages/pg-delta/src/core/objects/publication/changes/publication.alter.test.ts
@@ -156,6 +156,10 @@ describe("publication.alter", () => {
       stableId.table("public", "logs"),
       stableId.table("audit", "events"),
     ]);
+    expect(change.drops).toEqual([
+      stableId.table("public", "logs"),
+      stableId.table("audit", "events"),
+    ]);
     await assertValidSql(change.serialize());
     expect(change.serialize()).toBe(
       "ALTER PUBLICATION pub_drop_tables DROP TABLE public.logs, audit.events",
--- a/packages/pg-delta/src/core/objects/publication/changes/publication.alter.ts
+++ b/packages/pg-delta/src/core/objects/publication/changes/publication.alter.ts
@@ -151,6 +151,8 @@ export class AlterPublicationDropTables extends AlterPublicationChange {
   }
 
   get drops() {
+    // Treat ALTER PUBLICATION ... DROP TABLE as a destructive change so it runs
+    // in the drop phase before DROP TABLE removes the referenced relation.
     return this.tables.map((table) => stableId.table(table.schema, table.name));
   }
 
--- a/packages/pg-delta/src/core/objects/publication/changes/publication.alter.ts
+++ b/packages/pg-delta/src/core/objects/publication/changes/publication.alter.ts
@@ -150,6 +150,10 @@ export class AlterPublicationDropTables extends AlterPublicationChange {
     return Array.from(dependencies);
   }
 
+  get drops() {
+    return this.tables.map((table) => stableId.table(table.schema, table.name));
+  }
+
   serialize(): string {
     const targets = this.tables.map((table) => `${table.schema}.${table.name}`);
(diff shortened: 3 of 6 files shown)


jgoux (approved):
Great work on this!

PR #199, opened 2026-04-16
chore(tests): upgrade supabase postgres version to latest

avallete:
## Summary

This PR upgrades the Supabase-backed test containers to the latest image tags and changes our Supabase test bootstrap so it no longer assumes the raw `supabase/postgres` image already contains the full initialized schema.

To support that, it adds a maintainer workflow that generates a versioned SQL baseline for each supported Supabase Postgres major version. The script spins up a real `supabase start` project, diffs that fully initialized database against the matching bare `supabase/postgres` container, and checks in the resulting "full stack minus bare image" SQL fixture.

Our isolated Supabase tests now replay that baseline before running test-specific setup, so they start from the same service-managed state that a local Supabase stack would produce, without requiring us to boot the full Supabase stack for every test.

This branch also includes a few pg-delta correctness fixes:
- treat `etl` as a Supabase system schema
- preserve procedure comments when procedures are recreated
- handle PostgreSQL 17 generated-column transitions correctly, including recreating columns when switching between regular and generated states

## Why

Newer Supabase Postgres images no longer ship with the full set of service-owned migrations baked into the base image. Some of that setup now happens when other services initialize, which means a plain `supabase/postgres` container is no longer an accurate baseline for our tests.

We also do not want to replace our isolated-container test flow with `supabase start` for every test, since that would require bringing up the full stack and would make the suite much heavier and slower.

Instead, this PR captures the schema delta between the bare image and a fully initialized local Supabase stack once, versions it as SQL fixtures, and replays it during test bootstrap. That keeps the tests fast while aligning them with the real Supabase baseline.

## Test Plan

- add unit coverage for the new baseline sync helpers
- add an integration test proving the generated baseline exposes service-owned tables such as `storage.buckets`
- update the dbdev Supabase roundtrip test to build on top of the generated baseline
- add regression coverage for procedure comments, PostgreSQL 17 generated-column diffs, and Supabase `etl` filtering


packages/pg-delta/src/core/integrations/supabase.ts
```diff
@@ -16,6 +16,7 @@ const SUPABASE_SYSTEM_SCHEMAS = [
   "_supavisor",
   "auth",
   "cron",
+  "etl",
```
avallete, line 19: **note**

Use this PR to pass this update to the supabase integration, it's not needed as the schema will already be excluded due to role, but good to have it explicit here.

packages/pg-delta/src/core/objects/procedure/procedure.diff.test.ts
```diff
@@ -158,4 +159,28 @@ describe.concurrent("procedure.diff", () => {
...
+      { [branch.stableId]: branch },
+    );
+
+    expect(changes.some((change) => change instanceof CreateProcedure)).toBe(
+      true,
+    );
+    expect(
+      changes.some((change) => change instanceof CreateCommentOnProcedure),
+    ).toBe(true);
+  });
```
avallete, line 185: **note**

Doing the diff between base supabase image and "fully setup one" oncovered this bug on comment diffing for procedures.

packages/pg-delta/src/core/objects/table/table.diff.ts
```diff
@@ -745,10 +745,18 @@ export function diffTables(
...
+            if (
+              isGeneratedColumn &&
+              (isPostgresLowerThan17 || generatedStatusChanged)
+            ) {
               // For generated columns in < PostgreSQL 17, we need to drop and recreate
-              // instead of using SET EXPRESSION AS for computed columns
+              // instead of using SET EXPRESSION AS for computed columns. We also
+              // need to recreate the column when switching between regular and
+              // generated states because SET EXPRESSION only applies to existing
+              // generated columns.
```
avallete, line 759: **note**

Doing the diff between base supabase image and "fully setup one" uncovered a bug where using `SET EXPRESSION` on a non generated column doesn't set it's value. This is now fixed in the `table.alter.ts` as well.

jgoux (approved):

PR #212, opened 2026-04-21
test(pgdelta): Add tests for CHECK (FALSE) NO INHERIT constraints

### LINKED ISSUE #198: test(pg-delta): preserve column-less CHECK (FALSE) NO INHERIT on inheritance parents
## Context

Table-level CHECK constraints can reference no columns at all (for example `CHECK (FALSE)`) and can be declared `NO INHERIT`. On inheritance parents, that combination is sometimes used to block direct inserts on the parent while allowing child tables created with `INHERITS`.

pg-delta appears to model `no_inherit`, compare `check_expression`, and support `NO INHERIT` formatting, but there is still no integration regression that exercises the column-less CHECK + `NO INHERIT` + inherited-table path end to end. That leaves the empty-`conkey` extraction path unproven and risks silently weakening schema semantics if the constraint is lost or altered during diff/apply.

Related context: pgschema #386.

## Test Case to Reproduce

**Initial state (both databases):**

```sql
CREATE SCHEMA test_schema;
```

**Change to diff (branch only):**

```sql
CREATE TABLE test_schema.parent_base (
  id uuid PRIMARY KEY,
  name text NOT NULL,
  CONSTRAINT no_direct_insert CHECK (FALSE) NO INHERIT
);

CREATE TABLE test_schema.child (
  CONSTRAINT child_pkey PRIMARY KEY (id)
) INHERITS (test_schema.parent_base);
```

**Expected:** pg-delta preserves the `no_direct_insert` constraint, including both `CHECK (FALSE)` and `NO INHERIT`, when roundtripping or replaying the inherited-table schema.

**Actual:** there is currently no integration regression covering this exact scenario.

## Suggested Fix

- Add integration coverage in `packages/pg-delta/tests/integration/constraint-operations.test.ts` for `CHECK (FALSE) NO INHERIT` on a parent table.
- Add a second case covering `CREATE TABLE ... INHERITS (parent_base)` with the constraint present on the parent.
- Verify `packages/pg-delta/src/core/objects/table/table.model.ts` returns a stable empty array rather than `null` for `key_columns` on column-less CHECK constraints when `conkey` is empty.
- Re-run the constraint integration suite to ensure ordinary CHECK cases do not regress.

avallete:
## Summary
Added integration tests to verify proper handling of PostgreSQL `CHECK (FALSE) NO INHERIT` constraints in the constraint operations test suite.

## Key Changes
- Imported `expect` from `bun:test` to enable assertion statements
- Added test case for creating a `CHECK (FALSE) NO INHERIT` constraint on an inheritance parent table
- Added test case for creating a `CHECK (FALSE) NO INHERIT` constraint on a parent table with an inheriting child table
- Both tests verify that the generated SQL statements correctly include the `ADD CONSTRAINT ... CHECK (false) NO INHERIT` syntax
- The second test additionally verifies that the `INHERITS` clause is properly preserved in the generated statements

## Implementation Details
- Tests use the existing `roundtripFidelityTest` helper to validate schema roundtrip consistency
- Both test cases follow the established pattern of using `initialSetup`, `testSql`, and `assertSqlStatements`
- Assertions check for the presence of expected SQL fragments in the generated migration statements
- Tests cover both simple constraint creation and the more complex scenario with table inheritance

https://claude.ai/code/session_01M8bxUZzDHscoM728VFKb8f

Closes: https://github.com/supabase/pg-toolbelt/issues/198

### DIFF
--- a/packages/pg-delta/tests/integration/constraint-operations.test.ts
+++ b/packages/pg-delta/tests/integration/constraint-operations.test.ts
@@ -2,7 +2,7 @@
  * Integration tests for PostgreSQL constraint operations.
  */
 
-import { describe, test } from "bun:test";
+import { describe, expect, test } from "bun:test";
 import { POSTGRES_VERSIONS } from "../constants.ts";
 import { withDb, withDbIsolated } from "../utils.ts";
 import { roundtripFidelityTest } from "./roundtrip.ts";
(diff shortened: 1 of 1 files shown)


avallete (commented):
LGTM

jgoux (approved):

PR #214, opened 2026-04-21
fix(pgdelta): function signature changes to use DROP + CREATE instead of OR REPLACE

avallete:
## Summary
PostgreSQL's `CREATE OR REPLACE FUNCTION` has strict limitations on what can be changed in an existing function's signature. This PR fixes pg-delta to properly detect signature-breaking changes and emit `DROP FUNCTION` followed by `CREATE FUNCTION` instead of the unsupported `CREATE OR REPLACE` syntax.

## Key Changes

- **Signature-breaking field detection**: Introduced `SIGNATURE_BREAKING_FIELDS` to identify changes that PostgreSQL rejects in `CREATE OR REPLACE`, including:
  - Return type changes
  - Parameter type changes (affects `argument_types`)
  - Parameter name changes (PostgreSQL: "cannot change name of input parameter")
  - Parameter arity changes (affects `argument_count`)
  - Parameter default removal (PostgreSQL: "cannot remove parameter defaults")
  - Parameter mode changes

- **Separated alteration strategies**: Split non-alterable fields into two categories:
  - `SIGNATURE_BREAKING_FIELDS`: Require DROP + CREATE
  - `OR_REPLACEABLE_NON_ALTERABLE_FIELDS`: Safe for `CREATE OR REPLACE` (language, source_code, binary_path, sql_body)

- **Dependency cascading**: When a function signature changes, dependent objects (views, triggers, column defaults) are automatically replaced via the existing `expandReplaceDependencies` mechanism

- **Comment preservation**: When a signature change triggers DROP + CREATE, the function's comment is re-emitted since DROP destroys the old comment

## Implementation Details

- Modified `diffProcedures()` to check `signatureChanged` separately from other non-alterable changes
- Extracted procedure creation logic into `appendCreateProcedureChanges()` helper to avoid duplication when handling both created procedures and signature-changed procedures
- Added comprehensive test coverage for all signature change scenarios and dependent object cascading

https://claude.ai/code/session_014krHKmkYU65L8WWvgG5mXN


avallete (commented):
LGTM

jgoux (approved):

PR #216, opened 2026-04-22
feat(pg-delta): add security labels handling

avallete:
## Summary

Adds pg-delta support for PostgreSQL `SECURITY LABEL` extraction, diffing, serialization, and filtering across supported object types.

- Extracts and models security labels from `pg_seclabel` / `pg_shseclabel` for schemas, tables, columns, views, materialized views, sequences, routines, types, domains, event triggers, FDW objects, publications, subscriptions, and roles.
- Emits deterministic `SECURITY LABEL FOR ... IS ...` / `IS NULL` changes during diffing, with stable IDs and dependency ordering.
- Adds `scope: "security_label"` and provider-level filtering support in the filter DSL.
- Adds a prebuilt `dummy_seclabel` CI image flow so integration shards can test security labels without rebuilding the contrib image per shard.

## Notes

Role security labels are cluster-wide, so the role-label integration coverage uses isolated containers rather than database-only isolation. This avoids shared-container pollution between concurrent integration tests.

This PR includes changesets for the new security-label feature and the follow-up provider-filter fix.

## Test plan

- `PGDELTA_TEST_POSTGRES_VERSIONS=17 bun run test src/core/integrations/filter/dsl.test.ts`
- `PGDELTA_TEST_POSTGRES_VERSIONS=15,17,18 bun run test tests/integration/security-label-filter.test.ts tests/integration/security-label-operations.test.ts`
- Targeted Biome check on edited files
- `bun run check-types`

## Additional context

Closes: CLI-1386


avallete (commented):
LGTM

jgoux (approved):
Great work!

PR #223, opened 2026-04-23
fix(pg-delta): skip indexes when pg_get_indexdef() returns NULL

avallete:
## Summary
Modified the index extraction logic to gracefully handle cases where `pg_get_indexdef()` returns NULL instead of crashing with a ZodError. Indexes with NULL definitions are now filtered out with a debug log, allowing catalog extraction to continue.

## Changes
- **Schema validation**: Extended `indexPropsSchema` with a new `indexRowSchema` that allows `definition` to be nullable, with detailed comments explaining why NULL values can occur
- **Filtering logic**: Added a filter step after validation that removes rows with NULL definitions and logs them at debug level (`pg-delta:extract:index`) for troubleshooting
- **Test coverage**: Added comprehensive test suite covering:
  - Skipping rows with NULL definitions while processing valid indexes
  - Handling cases where all rows have NULL definitions without throwing errors
  - Returning all indexes when all definitions are valid

## Implementation Details
The `pg_get_indexdef()` function can return NULL under several conditions:
- Race conditions with concurrent DDL (e.g., index being dropped during extraction)
- Transient catalog metadata inconsistencies
- Recovery edge cases
- Role visibility issues

Rather than failing the entire catalog extraction, these unreadable indexes are now silently filtered out with debug logging, making the extraction process more resilient to transient PostgreSQL state issues.

https://claude.ai/code/session_01JepRL8RqbS6cCvnTYuRd12

### DIFF
--- a/.changeset/fix-index-null-definition.md
+++ b/.changeset/fix-index-null-definition.md
@@ -1,0 +1,5 @@
+---
+"@supabase/pg-delta": patch
+---
+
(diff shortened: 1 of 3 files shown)


avallete:
LGTM

jgoux (approved):

PR #224, opened 2026-04-23
fix(pg-delta): handle malformed percent-encoding in connection URLs

avallete:
## Summary

Fixes `URIError: URI malformed` crash in `poolConfigFromUrl` (Sentry [SUPABASE-API-7TV](https://supabase.sentry.io/issues/SUPABASE-API-7TV), 20+ events on `@supabase/pg-delta@1.0.0-alpha.16`). Connection URLs whose userinfo, database, or hostname contain a bare `%` or a truncated escape like `%x` crashed `createManagedPool` synchronously before any connection attempt.

## Fix

- Introduce `safeDecodeURIComponent` in `packages/pg-delta/src/core/connection-url.ts` — wraps `decodeURIComponent` and returns the input unchanged on `URIError` (rethrows anything else).
- Apply it at the three call sites in `poolConfigFromUrl` (`postgres-config.ts:343-346`) and the one in `normalizeConnectionUrl` (`connection-url.ts:65`).
- Non-IPv6 URLs still go through `pg-connection-string` (which has its own `%` workaround), unchanged.

## Reproduction

Full production stack trace reproduced locally with a percent-encoded IPv6 host + bare `%` password:

```
URIError: URI error
  at decodeURIComponent (unknown)
  at poolConfigFromUrl (postgres-config.ts:344:42)
  at createManagedPool (postgres-config.ts:389:28)
```

After the fix, the same invocation proceeds past URL parsing and surfaces a normal network/auth outcome instead.

## Test plan

- [x] 4 regression tests in `postgres-config.test.ts` (`malformed percent-encoding in userinfo/database`) flip from red to green
- [x] Existing `"host bracket strip survives percent-decoded username/password"` still green
- [x] 5 new `safeDecodeURIComponent` unit tests
- [x] 1 new `normalizeConnectionUrl` regression (hostname `%3A` + bare `%`)
- [x] Live end-to-end reproduction against a Supabase IPv6 URL: `URIError` before fix, normal network outcome after
- [x] `bun run format-and-lint`, `bun run check-types`, `bun run knip` all clean

https://claude.ai/code/session_015VxiD6vQzGnvVXKE3zMpTH

### DIFF
--- a/.changeset/fix-uri-error-poolconfig-malformed-percent.md
+++ b/.changeset/fix-uri-error-poolconfig-malformed-percent.md
@@ -1,0 +1,9 @@
+---
+"@supabase/pg-delta": patch
+---
+
+fix(pg-delta): handle malformed percent-encoding in connection URLs
+
+`poolConfigFromUrl` and `normalizeConnectionUrl` used `decodeURIComponent` directly on the URL's userinfo, database name, and hostname. A bare `%` or a truncated escape like `%x` in any of those fields raised `URIError: URI malformed`, crashing `createManagedPool` before a connection attempt was made (Sentry `SUPABASE-API-7TV`).
+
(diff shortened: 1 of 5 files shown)


avallete:
This should be an upstream concern not pg-delta one.

Issue #230, opened 2026-04-27
pg-delta: incorrect DROP ordering for dependent objects


Zohix1:
# Bug report

Hi! I've identified a pattern in pg-delta where DROP operations fail because they're generated in the wrong dependency order. The tool attempts to drop depended-upon objects (functions, types, sequences) before removing the dependencies that reference them, resulting in PostgreSQL errors like "cannot drop ... because other objects depend on it" (error 2BP01).

This affects multiple object types including functions with RLS policies, custom types used by columns, and sequences referenced by column defaults. I have reproducible test cases for all three scenarios.

---

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

### Describe the bug

pg-delta generates DROP operations in incorrect dependency order. It attempts to drop depended-upon objects (functions, types, sequences) before removing the dependencies that reference them, causing PostgreSQL errors:

```
ERROR 2BP01: cannot drop ... because other objects depend on it
```

This affects three categories of objects:
1. **Functions** with RLS POLICY dependencies
2. **Types** with column dependencies
3. **Sequences** with column DEFAULT dependencies

The pattern is consistent: dependency removal should occur **before** the object DROP, but pg-delta generates them **after**.

#### Case 1: DROP Function with RLS POLICY Dependencies

**Schema in both databases:**
```sql
CREATE FUNCTION check_access(user_id UUID) 
RETURNS BOOLEAN AS $$
BEGIN 
  RETURN TRUE; 
END;
$$ LANGUAGE plpgsql;

CREATE TABLE docs (id INT, owner_id UUID, content TEXT);
ALTER TABLE docs ENABLE ROW LEVEL SECURITY;

CREATE POLICY docs_policy ON docs
  USING (check_access(owner_id));
```

**Change to target database (function signature change):**
```sql
-- Remove old function and recreate with different parameters
DROP POLICY docs_policy ON docs;
DROP FUNCTION check_access(UUID);

CREATE FUNCTION check_access(user_id UUID, resource_id INT) 
RETURNS BOOLEAN AS $$
BEGIN 
  RETURN TRUE; 
END;
$$ LANGUAGE plpgsql;

CREATE POLICY docs_policy ON docs 
  USING (check_access(owner_id, id));
```

**Generated SQL by pg-delta:**
```sql
[...]

PR #232, opened 2026-04-28
test(pg-delta): add failing regressions for issue #230 DROP ordering bugs

avallete:
Three fidelity regressions covering the cases from issue #230 where
pg-delta emitted DROP statements before the ALTER that releases the
pg_depend edge to the dropped object, causing PostgreSQL to abort the
migration with error 2BP01:

1. tests/integration/alter-table-operations.test.ts unskips the existing
   `change column type from enum to text preserves default` test that was
   parked behind a `test.skip` for CLI-754.
2. tests/integration/sequence-operations.test.ts adds
   `drop sequence referenced by column default` covering the
   `ALTER COLUMN DROP DEFAULT` -> `DROP SEQUENCE` ordering case.
3. tests/integration/rls-operations.test.ts adds
   `replace function signature referenced by RLS policy` covering an
   RLS policy whose USING expression references a procedure whose
   signature changes in the same plan.

Each test fails on the parent commit by hitting `cannot drop ... because
other objects depend on it` (SQLSTATE 2BP01) when the generated plan is
applied, matching the failure modes reported in the issue.

https://claude.ai/code/session_01FRE6kdznpqGSfnd9r3uKWv


avallete (commented):
LGTM

jgoux (approved):

Issue #228, opened 2026-04-27
pg-delta: RENAME operations detected as DROP+CREATE


Zohix1:
# Bug report

Hi! I've been testing pg-delta with roundtrip schema verification and found an issue with how rename operations are handled. Table and column renames are being detected as DROP+CREATE instead of using native `ALTER ... RENAME` commands, which causes schema drift and breaks the roundtrip verification.

This leads to non-converging schemas and can be destructive in production migrations. 

---
- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

### Describe the bug

pg-delta detects table and column rename operations as DROP+CREATE sequences instead of using native `ALTER ... RENAME` commands. This causes three categories of schema drift:

1. **Sequence name mismatches** - When a table with SERIAL columns is renamed, the sequence is not renamed, breaking the roundtrip schema convergence test
2. **VIEW reference failures** - When columns are renamed via DROP+ADD, VIEW definitions referencing the old column name become invalid
3. **Cascading VIEW deletion** - When tables are renamed via DROP+CREATE, dependent VIEWs are dropped and recreated, causing additional schema changes on reapply

#### Case 1: Table Rename with SERIAL Sequence Name Mismatch

**Schema in both databases (source and target):**
```sql
CREATE TABLE old_table (
  id SERIAL PRIMARY KEY,
  name TEXT
);
-- This creates sequence: old_table_id_seq
-- Default value: nextval('old_table_id_seq'::regclass)
```

**Change applied to target database only:**
```sql
ALTER TABLE old_table RENAME TO new_table;
-- Sequence should be renamed: old_table_id_seq → new_table_id_seq
-- Default value should be: nextval('new_table_id_seq'::regclass)
```

**Run pg-delta to plan migration from source to target:**
```bash
pgdelta plan --source <source-url> --target <target-url> --format sql
```

**Actual generated SQL:**
```sql
DROP TABLE old_table CASCADE;
CREATE TABLE new_table (
  id INTEGER DEFAULT nextval('old_table_id_seq'::regclass),  -- Wrong sequence name
  name TEXT
);
```

**After applying to source database:**
- Roundtrip verification fails
- New table references old sequence name `old_table_id_seq`
- `pgdelta plan` still reports remaining changes
- Schema is not converged despite apply

[...]

PR #233, opened 2026-04-28
fix(pg-delta): add failing regression rename scenarios

avallete:
Three roundtripFidelityTest cases reproducing the failures reported in
https://github.com/supabase/pg-toolbelt/issues/228 against the current
diff path (without the accompanying fix):

1. Table rename with a SERIAL column. The plan emits DROP TABLE for the
   old name plus CREATE TABLE for the new name with the column default
   pointing at the original sequence; PostgreSQL cascade-drops the
   sequence with the old table, then CREATE TABLE fails with
   `relation "old_table_id_seq" does not exist`.
2. Column rename whose dependent view is also drop+recreated in the
   target. Today this scenario goes through the existing
   expand-replace machinery and succeeds, but it is pinned here so the
   companion table-rename fixes do not regress it.
3. Table rename with a dependent view (definition rewritten in branch).
   `DROP TABLE users` fails with `cannot drop table users because other
   objects depend on it` because the view is not promoted to drop+create
   when the table being dropped lacks a same-name create.

These tests fail (RED) at this commit and pass (GREEN) after the
follow-up fix.


avallete (commented):
LGTM

jgoux (approved):

PR #238, opened 2026-04-30
feat(pg-delta): retry catalog extractors when pg_get_*def returns NULL

avallete:
## Summary

This PR adds resilience to catalog extraction by implementing retry logic for cases where PostgreSQL's `pg_get_*def()` functions transiently return NULL. These functions can return NULL when the underlying catalog row is dropped concurrently or the catalog state is in flux during recovery. Previously, such rows were silently dropped after a single attempt; now the extraction process retries a configurable number of times before giving up.

## Key Changes

- **New `extract-with-retry.ts` module**: Implements `extractWithDefinitionRetry()` helper that retries queries when rows have NULL definitions, with configurable retry count and backoff strategy. Includes `resolveExtractRetries()` to read retry configuration from options or `PGDELTA_EXTRACT_RETRIES` environment variable.

- **Updated extractors**: Modified `extractIndexes()`, `extractTables()`, `extractProcedures()`, `extractTriggers()`, `extractRules()`, `extractViews()`, and `extractMaterializedViews()` to use the retry helper and accept `ExtractRetryOptions`.

- **Schema updates**: Extended Zod schemas to allow nullable `definition` fields at the row level (e.g., `indexRowSchema`, `procedureRowSchema`, `tableConstraintRowSchema`), enabling graceful handling of NULL definitions rather than validation errors.

- **Catalog extraction integration**: Updated `extractCatalog()` to accept `ExtractCatalogOptions` and pass retry configuration to all extractors.

- **Plan creation integration**: Updated `createPlan()` to pass `extractRetries` option from plan options to `extractCatalog()`.

- **Comprehensive test coverage**: Added test suites for all affected extractors covering:
  - Skipping rows with NULL definitions after retries exhausted
  - Handling cases where all rows have NULL definitions
  - Returning all rows when definitions are valid
  - Recovering from transient NULL definitions on retry

## Implementation Details

- Retry logic uses linear backoff: `backoffMs * attemptNumber`
- Default retry count is 1 (2 total attempts), configurable via `PGDELTA_EXTRACT_RETRIES` env var or options
- Rows with NULL definitions are filtered out only after all retries are exhausted, preventing data loss from transient issues
- The retry helper is generic and works with both flat schemas (definition on row) and nested schemas (definition on child collections like table constraints)

https://claude.ai/code/session_01DyaEfHPJ63hwzHbb3k8wBA


avallete (commented):
LGTM

jgoux (approved):

PR #239, opened 2026-05-01
ci: gate release-PR test skip on a path whitelist

avallete:
## Summary
Replace the hardcoded bot user check with a more robust detection mechanism for changesets release PRs. This allows the CI to skip heavy tests when only release artifacts (CHANGELOGs, package.json version bumps, and .changeset state) have changed, while still running the full test suite if any unexpected files are modified.

## Key Changes
- **Added `is-release-pr-content` output** to the `detect-changes` action that identifies when all changed files are part of the release-PR whitelist (CHANGELOGs, package.json version bumps, .changeset state)
- **Implemented file filtering logic** using `dorny/paths-filter` with a negation-based approach to detect any files outside the release-PR whitelist
- **Introduced `is-release-pr` output** in the test workflow that combines:
  - Event type check (pull_request only)
  - Branch check (changeset-release/main)
  - Content check (only release artifacts changed)
- **Replaced all bot user checks** throughout the workflow with the new `is-release-pr` output, making the detection more maintainable and less fragile

## Implementation Details
- The release-PR content detection uses a "predicate-quantifier: every" filter that matches files only when they satisfy ALL patterns, including negated ones. This fires on any file NOT in the release-PR artifacts list
- The whitelist includes:
  - `.changeset/pre.json`
  - `.changeset/*.md`
  - `packages/pg-delta/package.json` and `CHANGELOG.md`
  - `packages/pg-topo/package.json` and `CHANGELOG.md`
- All test jobs (type checking, linting, knip, pg-delta tests, pg-topo tests) now use the centralized `is-release-pr` condition instead of checking the bot user directly

https://claude.ai/code/session_01F6RDgU3axtTj8aHHuwAbTr


.github/actions/detect-changes/action.yml
```diff
@@ -17,6 +17,13 @@ outputs:
   root:
     description: Whether root config changed
     value: ${{ steps.filter.outputs.root }}
+  is-release-pr-content:
+    description: >-
+      'true' iff every changed file is part of the changesets release-PR
```
jgoux:
```suggestion
      'true' if every changed file is part of the changesets release-PR
```

.github/actions/detect-changes/action.yml
```diff
@@ -38,3 +45,30 @@ runs:
...
+    - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
+      id: non-release-files
+      with:
+        predicate-quantifier: "every"
+        filters: |
+          non-release-files:
+            - "**"
+            - "!.changeset/pre.json"
+            - "!.changeset/*.md"
+            - "!packages/pg-delta/package.json"
```
jgoux, line 62: Isn't excluding `package.json` going to skip tests when dependabot update the dependencies?
Or is it protected by the lock file? 🤔

jgoux (approved):

.github/actions/detect-changes/action.yml
avallete: Dependabot is safe — there are two things keeping it out of the fast-path:

1. **`bun.lock` is not in the whitelist.** Any dependency bump touches the lockfile, so `non-release-files` fires and `is-release-pr-content` flips to `'false'`. Tests run.
2. **Belt-and-suspenders: `head_ref == 'changeset-release/main'` gate** in `tests.yml`. Dependabot opens PRs on branches like `dependabot/npm_and_yarn/...`, never `changeset-release/main`, so even if a hypothetical "lockfile-less" update happened, `is-release-pr` would still be `'false'`.

So the only PRs that can claim the skip are: opened from `changeset-release/main` AND every changed file is in `{changeset state, packages/*/{package.json, CHANGELOG.md}}`. Anything else (dependabot, a human pushing onto the release branch, the bot misbehaving) runs the full suite.

---

PR #253, opened 2026-05-09
ci: cache pinned Bun toolchain and add retried install fallback

avallete:
## Why

CI has been flaking on transient 502s from the GitHub release CDN when `oven-sh/setup-bun@v2` downloads Bun:

```
Downloading a new version of Bun: https://github.com/oven-sh/bun/releases/download/bun-v1.3.13/bun-linux-x64.zip
Unexpected HTTP response: 502
Waiting 13 seconds before trying again
Unexpected HTTP response: 502
Waiting 20 seconds before trying again
Error: Error: Unexpected HTTP response: 502
```

`setup-bun` only retries the download twice. Every job in `tests.yml` and `release.yml` runs `./.github/actions/setup`, and the integration matrix alone is 24 shards, so a brief CDN hiccup takes the whole pipeline down. The setup also re-downloaded Bun on every job because:

- `bun-version: latest` has no stable cache key.
- Hosted runners' tool cache (`/opt/hostedtoolcache/bun/...`) is wiped per-runner.
- `setup-bun`'s `no-cache: false` only caches `~/.bun/install/cache` (the dependency install cache), not the Bun binary.

## What

Single file edit: [.github/actions/setup/action.yml](.github/actions/setup/action.yml).

1. **Pin Bun** to `1.3.13` (what `latest` resolves to today) via a `BUN_VERSION` env var written to `$GITHUB_ENV` in the first step. Bumping is a one-line change and automatically invalidates the cache.
2. **Cache `/opt/hostedtoolcache/bun`** with `actions/cache@v4`, keyed on `os/arch/BUN_VERSION`. On cache hit, `setup-bun` finds the binary already on disk and skips the GitHub release CDN download entirely.
3. **Retried fallback** — keep `oven-sh/setup-bun@v2` as the primary install (with `continue-on-error: true`) and add a `nick-fields/retry@v3` step that runs `curl -fsSL https://bun.sh/install | bash -s "bun-v${BUN_VERSION}"` with up to 5 attempts and a 30s backoff. The official installer hits `bun.sh`, a different host than the GitHub release CDN, so it's unlikely to be impacted by the same incident.
4. A final `bun --version` step makes any failure fast and obvious if neither path produced a working `bun` on PATH.
5. Adds the previously missing `description:` field to the action metadata to clear an `actionlint` warning.

## Flow

```mermaid
flowchart TD
    Start[Setup composite] --> Restore["actions/cache@v4 restore<br/>/opt/hostedtoolcache/bun"]
    Restore --> Primary["oven-sh/setup-bun v2<br/>continue-on-error: true"]
    Primary -->|cache hit, no download| Verify
    Primary -->|cache miss, success| Verify
    Primary -->|cache miss, 502 failure| Fallback["Fallback install<br/>nick-fields/retry v3<br/>curl bun.sh/install"]
    Fallback --> Verify[bun --version sanity check]
    Verify --> Node[actions/setup-node v6]
    Node --> Install[bun install]
```

## Validation

- `actionlint` clean against all workflows and the composite action (also fixes a pre-existing missing-`description` warning).
- The first run on this branch will be a cold cache (exercises `setup-bun` and writes the cache). Subsequent runs on the branch should show `Restore Bun toolchain cache` reporting a cache hit and `Install Bun` skipping the download in its log.

## Out of scope

- Adding similar retry/cache to `actions/setup-node@v6`. Not reported as flaky and it has its own caching/retry behavior.
- Caching `~/.bun/install/cache` (Bun's package install cache). `setup-bun` already handles that with its default `no-cache: false`.
- Bumping Bun beyond what `latest` currently resolves to. Pinning to `1.3.13` preserves current behavior; future bumps are a separate, intentional change.
- Changeset entry — this only touches CI infrastructure under `.github/`, not package behavior.


avallete:
@copilot resolve the merge conflicts in this pull request pin all actions with commit hash for security compliance.

jgoux (approved):

PR #264, opened 2026-05-25
feat(pg-delta): add --filter to catalog-export

leandrocp:
This is a proposal to add `--filter` to `catalog-export`.

My use case is to export the `realtime` schema which is used in an internal admin tool to fix drifted schemas, see https://github.com/supabase/realtime/pull/1866 (already deployed to PROD).

That page currently works because `plan` accepts `--filter` but the baseline (catalog-export) ends with [way more data than it needs](https://github.com/supabase/realtime/blob/20e818e911af30b80190aec9988d05495b4484c7/priv/repo/tenant_db_baseline.json).

Disclaimer: I used Claude to assist the development and I don't know much about the internals of this project, but I tested on realtime and the generated catalog looks good.

### DIFF
--- a/packages/pg-delta/src/cli/commands/catalog-export.ts
+++ b/packages/pg-delta/src/cli/commands/catalog-export.ts
@@ -35,7 +35,7 @@ export const catalogExportCommand = buildCommand({
       filter: {
         kind: "parsed",
         brief:
-          'Filter DSL as inline JSON to filter changes (e.g., \'{"schema":"public"}\').',
+          'Filter DSL as inline JSON to filter changes (e.g., \'{"*/schema": "app"}\').',
         parse: (value: string): FilterDSL => {
           try {
             return JSON.parse(value) as FilterDSL;
--- a/packages/pg-delta/src/cli/commands/catalog-export.ts
+++ b/packages/pg-delta/src/cli/commands/catalog-export.ts
@@ -4,11 +4,13 @@
 
 import { writeFile } from "node:fs/promises";
 import { buildCommand, type CommandContext } from "@stricli/core";
+import { filterCatalog } from "../../core/catalog.filter.ts";
 import { extractCatalog } from "../../core/catalog.model.ts";
 import {
   serializeCatalog,
   stringifyCatalogSnapshot,
 } from "../../core/catalog.snapshot.ts";
+import type { FilterDSL } from "../../core/integrations/filter/dsl.ts";
 import { createManagedPool } from "../../core/postgres-config.ts";
 
 export const catalogExportCommand = buildCommand({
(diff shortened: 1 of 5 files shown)


packages/pg-delta/src/core/catalog.filter.ts
```diff
@@ -0,0 +1,84 @@
...
+import { Catalog, createEmptyCatalog } from "./catalog.model.ts";
+import { compileFilterDSL, type FilterDSL } from "./integrations/filter/dsl.ts";
+
+export async function filterCatalog(
+  catalog: Catalog,
+  filter: FilterDSL,
+): Promise<Catalog> {
+  const empty = await createEmptyCatalog(catalog.version, catalog.currentUser);
+  const changes = diffCatalogs(empty, catalog);
+  const filterFn = compileFilterDSL(filter);
```
avallete, line 38: Concretely: before this line, check `(filter as Record<string, unknown>).cascade === true` and throw — something like:

```ts
if (
  typeof filter === "object" &&
  filter !== null &&
  (filter as Record<string, unknown>).cascade === true
) {
  throw new Error(
    "Filter DSL `cascade: true` is not supported by catalog-export: " +
      "scoped snapshots are intentionally partial. Out-of-scope owners, " +
      "roles, and types must exist on the target DB at apply time.",
  );
}
```

(Composition patterns can also carry `cascade` on nested nodes — for now only the top-level field is read by `createPlan`, so matching that scope is fine.)

Another option might also be to just report it via a warning, mentioning that it is skipped / invalid for catalog filter

packages/pg-delta/src/cli/commands/catalog-export.ts
```diff
@@ -30,6 +32,21 @@ export const catalogExportCommand = buildCommand({
         parse: String,
         optional: true,
       },
+      filter: {
+        kind: "parsed",
+        brief:
```
avallete, line 37: Tiny nit while we're here: the example in this `brief` (`'{"schema":"public"}'`) doesn't actually match anything — the DSL paths are `*/schema`, `table/schema`, etc. The tests use `{"*/schema": "app"}`. Worth updating the example to the working form so users don't copy-paste a no-op.

---

leandrocp: Good point, updated.

packages/pg-delta/src/core/catalog.filter.ts
leandrocp: Since I don't know much about the project internals, I just applied your suggestion and added a test, but happy to make more changes if needed.

avallete (approved):

PR #265, opened 2026-06-03
fix(pg-delta): clear connect timeout

leandrocp:
The connect process hungs for the duration of PGDELTA_CONNECT_TIMEOUT_MS.

I noticed this issue when I tried to increase the timeout on the [Realtime admin page](https://github.com/supabase/realtime/blob/main/lib/realtime_web/dashboard/tenant_migrations.ex) we use to fix drifted schemas using pgdelta. Increasing to 60s was causing the page to hang for 60s even tho the connection was taking a couple seconds to finish locally.

### DIFF
--- a/.changeset/fix-connect-timeout-timer-leak.md
+++ b/.changeset/fix-connect-timeout-timer-leak.md
@@ -1,0 +1,7 @@
+---
+"@supabase/pg-delta": patch
+---
+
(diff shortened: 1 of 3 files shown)


avallete (approved):
LGTM !

Thank's for your contribution !

PR #266, opened 2026-06-03
fix(pg-delta): stop re-validating NOT VALID constraints

leandrocp:
Do not place a `VALIDATE CONSTRAINT` after a `ADD CONSTRAINT foo CHECK (...) NOT VALID` because the first already has the validation info and revalidating cause the planner to never converge to the final state.

In Realtime we [added a NOT VALID constraint](https://github.com/supabase/realtime/blob/e5ac7e233f40a6e761d3696ab12dd157a993a628/lib/realtime/tenants/repo/migrations/20260514120000_add_binary_payload_to_messages.ex#L10-L11) and now running pg-delta plan generates:

```
-- Risk: safe
ALTER TABLE realtime.messages DROP CONSTRAINT messages_payload_exclusive;

ALTER TABLE realtime.messages ADD CONSTRAINT messages_payload_exclusive CHECK (payload IS NULL OR binary_payload IS NULL) NOT VALID;

ALTER TABLE realtime.messages VALIDATE CONSTRAINT messages_payload_exclusive;
```

Which never converges because the last statement revert the expected state.

### DIFF
--- a/.changeset/feat-validate-constraint-shortcut.md
+++ b/.changeset/feat-validate-constraint-shortcut.md
@@ -1,0 +1,11 @@
+---
+"@supabase/pg-delta": patch
+---
+
+feat(pg-delta): emit `VALIDATE CONSTRAINT` shortcut when only `validated` flips from false to true
+
+When the only difference between main and branch for an existing table constraint is `convalidated` flipping from `false` to `true` (i.e. the user wants to validate a previously `NOT VALID` constraint), pg-delta now emits a single `ALTER TABLE ... VALIDATE CONSTRAINT ...` instead of dropping and re-adding the constraint.
+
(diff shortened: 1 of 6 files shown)


avallete (approved):
LGTM !

Only did an additional PR to fill one gap of the `NO VALIDATED -> VALIDATED` change.

Added an e2e test for it. And ensured that in such case, we "just" generate a `ALTER + VALIDATE` change instead of dropping + recreating the constraint entierly.

Thank's for your contribution !

leandrocp:
> LGTM !
> 
> Only did an additional PR to fill one gap of the `NO VALIDATED -> VALIDATED` change.
> 
> Added an e2e test for it. And ensured that in such case, we "just" generate a `ALTER + VALIDATE` change instead of dropping + recreating the constraint entierly.
> 
> Thank's for your contribution !

Thanks!

Issue #263, opened 2026-05-25
pg-delta: missing dependency chain resolution for ALTER COLUMN TYPE and DROP FUNCTION with Policy/View dependencies


Zohix1:
## Bug report

pg-delta fails to resolve dependency chains when generating `ALTER COLUMN TYPE` or `DROP FUNCTION` statements. It does not handle Policy or View dependencies that block these operations, resulting in PostgreSQL errors `0A000` (cannot alter type of a column used in a policy/view) and `2BP01` (cannot drop function because other objects depend on it).

This is a systematic issue affecting any schema where RLS Policies or Views reference the column being altered or the function being dropped. I have reproducible test cases for all scenarios.

- I confirm this is a bug with Supabase, not with my own application.
- I confirm I have searched the Docs, GitHub Discussions, and Discord.

## Describe the bug

pg-delta generates DDL that violates PostgreSQL's dependency rules in three related ways:

1. **0A000 (Policy)**: `ALTER COLUMN TYPE` is emitted without first dropping Policies that reference the column — pg-delta uses `ALTER POLICY` instead of `DROP POLICY`, which doesn't resolve the dependency
2. **0A000 (View)**: `ALTER COLUMN TYPE` is emitted before `DROP VIEW` — pg-delta generates the right statements but in the wrong order
3. **2BP01**: `DROP FUNCTION` is emitted without first dropping Policies that reference the function

All three errors share the same root cause: pg-delta does not properly resolve dependency metadata (`pg_depend` / `pg_policy` / `pg_rewrite`) before generating DDL, and therefore cannot produce the correct drop-alter-recreate sequence.

### Case 1: ALTER COLUMN TYPE blocked by RLS Policy (0A000)

Schema in both databases:

```sql
CREATE TYPE public.user_role_enum AS ENUM ('admin', 'user', 'guest');

CREATE TABLE public.solution_categories_with_policy (
  id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
  name text NOT NULL,
  role text NOT NULL
);

ALTER TABLE public.solution_categories_with_policy ENABLE ROW LEVEL SECURITY;

CREATE POLICY "categories_admin_manage" ON public.solution_categories_with_policy
  FOR ALL
  TO public
  USING (role = 'admin')
  WITH CHECK (role = 'admin');
```

Change to target database (column type change from `text` to `user_role_enum`):

```sql
DROP POLICY "categories_admin_manage" ON public.solution_categories_with_policy;
[...]

PR #272, opened 2026-06-06
fix(pg-delta): recreate policies for replaced function dependencies

xmh1011:
## Summary

Refs #263.

This PR fixes the first, isolated failure from #263: replacing a function signature that is referenced by an RLS policy. The old plan dropped the function before recreating the dependent policy, so PostgreSQL rejected the migration with `2BP01` because the existing policy still depended on the old function signature.

The catalog already extracts the authoritative dependency from `pg_depend` into `catalog.depends`; the missing piece was that `expandReplaceDependencies` did not know how to promote RLS policies into the replacement cascade. This change adds RLS policy handling there, so a dependent policy is dropped before the old function and recreated after the new function. Unchanged policy comments are recreated too, matching the existing comment-preservation behavior for other replaced objects.

This intentionally does not address the column rewrite / view ordering cases from #263. Those need a separate follow-up so the replace cascade and in-place column alter ordering can be handled without widening this PR.

## RED evidence

Manual reproduction of the function/policy case on current `main` generated only:

- `DROP FUNCTION public.alter_function_sign_policy_dependent_check_role(uuid, text)`
- `CREATE FUNCTION public.alter_function_sign_policy_dependent_check_role(uuid, text, text DEFAULT 'default'::text)`

Applying that plan failed with:

```text
2BP01: cannot drop function public.alter_function_sign_policy_dependent_check_role(uuid,text) because other objects depend on it
detail: policy alter_function_sign_policy_dependent_check_role_policy on table public.alter_function_sign_policy_dependent_profiles depends on function public.alter_function_sign_policy_dependent_check_role(uuid,text)
```

The new integration regression also failed before the fix because no `DROP POLICY` statement was generated (`dropPolicyIdx === -1`).

## Validation

- `PGDELTA_SKIP_DUMMY_SECLABEL_BUILD=1 PGDELTA_TEST_POSTGRES_VERSIONS=17 TESTCONTAINERS_RYUK_DISABLED=true ../../node_modules/.bin/bun run test src/core/expand-replace-dependencies.test.ts`
- `PGDELTA_SKIP_DUMMY_SECLABEL_BUILD=1 PGDELTA_TEST_POSTGRES_VERSIONS=17 TESTCONTAINERS_RYUK_DISABLED=true ../../node_modules/.bin/bun run test --test-name-pattern "policy depending on a replaced function" tests/integration/policy-dependencies.test.ts`
- `PGDELTA_SKIP_DUMMY_SECLABEL_BUILD=1 PGDELTA_TEST_POSTGRES_VERSIONS=17 TESTCONTAINERS_RYUK_DISABLED=true ../../node_modules/.bin/bun run test tests/integration/policy-dependencies.test.ts`
- `./node_modules/.bin/bun run format-and-lint:fix`
- `./node_modules/.bin/bun run check-types`
- `./node_modules/.bin/bun run knip --fix`
- `PGDELTA_SKIP_DUMMY_SECLABEL_BUILD=1 PGDELTA_TEST_POSTGRES_VERSIONS=17 TESTCONTAINERS_RYUK_DISABLED=true ./node_modules/.bin/bun run test:pg-delta`
  - Result: 1605 pass, 24 skip, 1 todo, 1 fail. The single failure was `dbdev declarative roundtrip (pg15)` with a PostgreSQL connection timeout.
- Focused rerun of the failed case after explicitly pulling the missing local image `postgres:15.14-alpine`:
  - `PGDELTA_SKIP_DUMMY_SECLABEL_BUILD=1 PGDELTA_TEST_POSTGRES_VERSIONS=15 TESTCONTAINERS_RYUK_DISABLED=true ../../node_modules/.bin/bun run test --test-name-pattern "exported schema roundtrips to 0 remaining changes with supabase integration" tests/integration/dbdev-roundtrip.test.ts`
  - Result: 1 pass, 0 fail.

`PGDELTA_SKIP_DUMMY_SECLABEL_BUILD=1` was used only for local sandbox execution, so security-label-specific tests were skipped locally as documented in `AGENTS.md`.

### DIFF
--- a/.changeset/kind-times-deny.md
+++ b/.changeset/kind-times-deny.md
@@ -1,0 +1,5 @@
+---
+"@supabase/pg-delta": patch
+---
+
+Recreate RLS policies that depend on replaced functions.
--- a/packages/pg-delta/src/core/expand-replace-dependencies.test.ts
+++ b/packages/pg-delta/src/core/expand-replace-dependencies.test.ts
@@ -6,6 +6,10 @@ import { DefaultPrivilegeState } from "./objects/base.default-privileges.ts";
 import { CreateProcedure } from "./objects/procedure/changes/procedure.create.ts";
 import { DropProcedure } from "./objects/procedure/changes/procedure.drop.ts";
 import { Procedure } from "./objects/procedure/procedure.model.ts";
+import { CreateCommentOnRlsPolicy } from "./objects/rls-policy/changes/rls-policy.comment.ts";
+import { CreateRlsPolicy } from "./objects/rls-policy/changes/rls-policy.create.ts";
+import { DropRlsPolicy } from "./objects/rls-policy/changes/rls-policy.drop.ts";
+import { RlsPolicy } from "./objects/rls-policy/rls-policy.model.ts";
 import { CreateSequence } from "./objects/sequence/changes/sequence.create.ts";
 import { DropSequence } from "./objects/sequence/changes/sequence.drop.ts";
 import { diffSequences } from "./objects/sequence/sequence.diff.ts";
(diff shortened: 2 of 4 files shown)


avallete (approved):
LGTM thank's for your contribution!

PR #275, opened 2026-06-06
fix(pg-delta): recreate policies for column rewrite dependencies

xmh1011:
Refs #263

This is the third incremental fix for issue #263, scoped to Case 1: `ALTER TABLE ... ALTER COLUMN ... TYPE` fails when an existing RLS policy still depends on the rewritten column.

## What changed

For this case, pg-delta was producing the right table rewrite but the wrong policy handling:

1. `diffRlsPolicies` saw the policy expression text change after the column type changed from `text` to an enum, so it emitted `ALTER POLICY ... USING` and `ALTER POLICY ... WITH CHECK`.
2. PostgreSQL does not allow `ALTER COLUMN TYPE` while the old policy definition still depends on that column, so the policy must be fully dropped before the rewrite and recreated after it.
3. The table rewrite is an in-place mutation. After #273, these changes expose the affected stable ids through `BaseChange.invalidates` rather than adding concrete subtype checks to generic expansion/sorting code.

This PR now builds on #273's `invalidates` primitive:

- collect invalidated stable ids from `change.invalidates`;
- use `mainCatalog.depends` to find surviving `rlsPolicy:*` objects whose policy expressions depend on those ids;
- add `DropRlsPolicy` / `CreateRlsPolicy` for those policies;
- remove same-policy `ALTER POLICY` changes that are superseded by the replacement.

The fix stays within existing pg_catalog/pg_depend metadata and does not parse policy SQL or couple `expandReplaceDependencies` to `AlterTableAlterColumnType`.

## RED evidence

- Initial focused unit regression failed because `expandReplaceDependencies` did not add `DropRlsPolicy` for a policy whose `pg_depend` row referenced the rewritten column.
- Initial focused PG17 integration regression failed because generated SQL had `ALTER TABLE ... ALTER COLUMN role TYPE ...` followed by `ALTER POLICY ... USING` / `ALTER POLICY ... WITH CHECK`, with no `DROP POLICY` / `CREATE POLICY` around the column rewrite.
- Manual apply reproduced PostgreSQL `0A000: cannot alter type of a column used in a policy definition`.
- After rebasing onto #273, I added a design regression using a generic invalidating change. RED before the refactor: `expand-replace-dependencies.test.ts` failed with `Expected: true / Received: false` for `DropRlsPolicy` because the helper still checked `AlterTableAlterColumnType` directly.

## Local validation

- Unit-only focused invalidates regression: `1 pass / 0 fail`.
- Full `expand-replace-dependencies.test.ts`: `9 pass / 0 fail`.
- PG17 focused integration for `policy depending on a column type rewrite is dropped and recreated`: `1 pass / 0 fail`.
- Full PG17 `policy-dependencies.test.ts`: `9 pass / 0 fail` (one connection timeout retried by the test runner; final result was green).
- `./node_modules/.bin/bun run format-and-lint:fix`: passed.
- `./node_modules/.bin/bun run check-types`: passed.
- `./node_modules/.bin/bun run knip --fix`: passed; knip only reported existing configuration hints.

Local integration validation used `PGDELTA_SKIP_DUMMY_SECLABEL_BUILD=1` to avoid rebuilding the local dummy security-label image; CI should keep security-label coverage enabled.

### DIFF
--- a/packages/pg-delta/src/core/expand-replace-dependencies.test.ts
+++ b/packages/pg-delta/src/core/expand-replace-dependencies.test.ts
@@ -20,7 +20,6 @@ import { diffSequences } from "./objects/sequence/sequence.diff.ts";
 import { Sequence } from "./objects/sequence/sequence.model.ts";
 import {
   AlterTableAlterColumnSetDefault,
-  AlterTableAlterColumnType,
   AlterTableChangeOwner,
   AlterTableDropColumn,
   AlterTableDropConstraint,
@@ -49,12 +48,27 @@ function mockChange(overrides: {
     scope: "object",
     creates,
     drops,
+    invalidates: [],
     requires: [],
     table: { schema: "public", name: "t" },
     serialize: () => [],
     get requiresForDrop(): string[] {
       return [];
     },
+  } as unknown as Change;
+}
+
+function mockInvalidatingChange(invalidates: string[]): Change {
+  return {
+    objectType: "table",
+    operation: "alter",
+    scope: "object",
+    creates: [],
+    drops: [],
+    invalidates,
+    requires: [],
+    table: { schema: "public", name: "t" },
+    serialize: () => "",
   } as unknown as Change;
 }
 
@@ -826,7 +840,7 @@ describe("expandReplaceDependencies", () => {
     ).toBe(true);
   });
 
-  test("promotes dependent RLS policy when a referenced column is rewritten", async () => {
+  test("promotes dependent RLS policy when a referenced column is invalidated", async () => {
     const baseline = await createEmptyCatalog(170000, "postgres");
     const columnTemplate = {
       position: 1,
@@ -922,11 +936,9 @@ describe("expandReplaceDependencies", () => {
       withCheckExpression: branchPolicy.with_check_expression,
     });
     const changes: Change[] = [
-      new AlterTableAlterColumnType({
-        table: branchTable,
-        column: branchRoleColumn,
-        previousColumn: mainRoleColumn,
-      }),
+      mockInvalidatingChange([
+        "column:public.solution_categories_with_policy.role",
+      ]),
       alterUsing,
       alterWithCheck,
     ];
--- a/packages/pg-delta/src/core/expand-replace-dependencies.test.ts
+++ b/packages/pg-delta/src/core/expand-replace-dependencies.test.ts
@@ -6,6 +6,10 @@ import { DefaultPrivilegeState } from "./objects/base.default-privileges.ts";
 import { CreateProcedure } from "./objects/procedure/changes/procedure.create.ts";
 import { DropProcedure } from "./objects/procedure/changes/procedure.drop.ts";
 import { Procedure } from "./objects/procedure/procedure.model.ts";
+import {
+  AlterRlsPolicySetUsingExpression,
+  AlterRlsPolicySetWithCheckExpression,
+} from "./objects/rls-policy/changes/rls-policy.alter.ts";
 import { CreateCommentOnRlsPolicy } from "./objects/rls-policy/changes/rls-policy.comment.ts";
 import { CreateRlsPolicy } from "./objects/rls-policy/changes/rls-policy.create.ts";
 import { DropRlsPolicy } from "./objects/rls-policy/changes/rls-policy.drop.ts";
@@ -16,6 +20,7 @@ import { diffSequences } from "./objects/sequence/sequence.diff.ts";
 import { Sequence } from "./objects/sequence/sequence.model.ts";
 import {
   AlterTableAlterColumnSetDefault,
+  AlterTableAlterColumnType,
   AlterTableChangeOwner,
   AlterTableDropColumn,
   AlterTableDropConstraint,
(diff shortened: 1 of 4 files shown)


xmh1011:
Follow-up after comparing this PR with the review discussion on #273.

This PR still addresses a separate issue #263 case: an RLS policy whose expression depends on a column being rewritten needs to be promoted from `ALTER POLICY` changes to a `DROP POLICY` + `CREATE POLICY` replacement. Otherwise PostgreSQL can still block `ALTER TABLE ... ALTER COLUMN ... TYPE` because the policy expression is bound to the old column/type definition.

One design note before this is finalized: #273 introduced the `BaseChange.invalidates` primitive so in-place rewrites can declare the stable ids they invalidate without adding concrete `instanceof` checks to generic planning/sorting code. Once #273 lands in `main`, I think this PR should be rebased and adjusted to consume `change.invalidates` instead of importing/checking `AlterTableAlterColumnType` inside `expand-replace-dependencies.ts`.

Planned cleanup after #273 merges:
- rebase this branch onto the updated `main`;
- rename the column-specific helper to something like `collectInvalidatedRlsPolicyReplacements`;
- collect invalidated stable ids from `change.invalidates` and use `pg_depend` to find dependent `rlsPolicy:*` objects;
- keep the current focused regression and integration coverage, with the unit test wording updated to reflect the `invalidates` primitive rather than a concrete column-rewrite subtype.

I am intentionally not stacking #273's infrastructure onto this PR before #273 merges, to keep this diff reviewable and avoid duplicating the same base changes across two open PRs.

xmh1011:
@avallete Follow-up after #273 merged:

I rebased this branch onto the current `main` (`9f01826`) and updated the implementation to consume the `BaseChange.invalidates` primitive from #273 instead of checking `AlterTableAlterColumnType` inside `expandReplaceDependencies`.

Pushed `e5b75ee`:
- `expandReplaceDependencies` now collects invalidated stable ids from `change.invalidates`;
- it uses `mainCatalog.depends` to find surviving dependent `rlsPolicy:*` objects and promotes those policies to `DROP POLICY` + `CREATE POLICY`;
- same-policy `ALTER POLICY` changes are still removed when superseded by that replacement;
- the unit regression now uses a generic invalidating change so it verifies the design boundary, not a concrete table-change subtype.

RED after the rebase, before the refactor: the generic invalidates regression failed with `Expected: true / Received: false` for `DropRlsPolicy` because the helper still checked `AlterTableAlterColumnType` directly.

Validation:
- focused invalidates unit regression: `1 pass / 0 fail`;
- full `expand-replace-dependencies.test.ts`: `9 pass / 0 fail`;
- PG17 focused policy integration: `1 pass / 0 fail`;
- full PG17 `policy-dependencies.test.ts`: `9 pass / 0 fail` (one connection timeout retried by the runner; final result green);
- `format-and-lint:fix`, `check-types`, and `knip --fix` passed; knip only reported existing configuration hints.

packages/pg-delta/src/core/expand-replace-dependencies.ts
```diff
@@ -281,11 +293,90 @@ export function expandReplaceDependencies({
...
+      branchCatalog,
+    );
+    if (!resolved || resolved.kind !== "rls_policy") continue;
+
+    const addDrop = !droppedIds.has(targetId);
+    const addCreate = !createdIds.has(targetId);
+    const replacementChanges = buildReplaceChanges(resolved, {
+      addDrop,
+      addCreate,
+    });
```
avallete, line 353: Nit: `diffContext` isn't forwarded here, unlike the `buildReplaceChanges(resolved, { addDrop, addCreate, diffContext })` call in the BFS walk below. It's harmless today since the `rls_policy` branch of `buildReplaceChanges` doesn't use it, but if policy replacement ever grows owner/comment/ACL replay (like views did with `buildCreateViewReplacementChanges`), this call site would silently miss it. Passing it through costs nothing and keeps the two call sites symmetric.

---

packages/pg-delta/src/core/expand-replace-dependencies.ts
```diff
@@ -281,11 +293,90 @@ export function expandReplaceDependencies({
...
+    }
+  }
+  if (invalidatedIds.size === 0) return [];
+
+  const replacements: Change[] = [];
+  for (const dep of mainCatalog.depends) {
+    if (!invalidatedIds.has(dep.referenced_stable_id)) continue;
+
+    const targetId = normalizeDependentId(dep.dependent_stable_id);
+    if (!targetId?.startsWith("rlsPolicy:")) continue;
```
avallete, line 337: Non-blocking, for the #263 tracking issue: this pass only promotes `rlsPolicy:*` dependents of invalidated ids. Other dependents bound to a rewritten column hit the same PostgreSQL wall — e.g. an *unchanged* view on the rewritten column still blocks `ALTER COLUMN TYPE` with `cannot alter type of a column used by a view or rule` (#273 only ordered view drops that already existed in the diff; it doesn't synthesize a drop for a view with no diff).

A future iteration could seed the invalidated ids into the existing BFS walk (as roots that traverse but don't get replaced themselves) instead of a kind-specific scan — that would reuse `resolveObjectForStableId`/`buildReplaceChanges` and pick up transitive dependents for free. Keeping this PR scoped to policies is fine; just worth a note on #263 so the remaining cases aren't lost.

---

packages/pg-delta/src/core/expand-replace-dependencies.test.ts
```diff
@@ -53,6 +58,20 @@ function mockChange(overrides: {
...
+  return {
+    objectType: "table",
+    operation: "alter",
+    scope: "object",
+    creates: [],
+    drops: [],
+    invalidates,
+    requires: [],
+    table: { schema: "public", name: "t" },
+    serialize: () => "",
```
avallete, line 71: Nit: the sibling `mockChange` helper uses `serialize: () => []` — cosmetic inconsistency only, but matching it keeps the two mocks interchangeable.

---

avallete (approved):
**Approve with minor notes.** I independently verified the RED→GREEN evidence and ran the suites locally on the PR head (rebased on current `main`, which includes #273).

## Verification performed

- **RED reproduced**: with `main`'s version of `expand-replace-dependencies.ts` and this PR's tests, the new unit regression fails with exactly the reported `Expected: true / Received: false` for `DropRlsPolicy`; the 8 pre-existing tests stay green in both states, so the fix changes only the intended behavior.
- **GREEN**: full `expand-replace-dependencies.test.ts` — 9 pass / 0 fail; focused PG17 integration test — 1 pass (snapshot pins `DROP POLICY → ALTER ... TYPE → CREATE POLICY`); full PG17 `policy-dependencies.test.ts` — 9 pass / 0 fail; `check-types` and format/lint clean.
- Note: only the `Lint Pull Request` check has run on the latest push — the test workflows still need a maintainer to approve the fork CI run.

## Correctness highlights

- Promotion is driven by `pg_depend` edges rather than the presence of `ALTER POLICY` changes, so a policy whose deparsed expression is **unchanged** (which still blocks the rewrite in PostgreSQL) is correctly promoted too — an "intercept the alters" approach would have missed that case.
- Dropped/created policies are correctly skipped: `resolveObjectForStableId` requires the policy in both catalogs, so a policy removed in branch keeps its existing `DropRlsPolicy` (ordered before the rewrite via #273's drop-phase `invalidates` folding), and a new branch policy is untouched.
- No double work: promoted ids feed `createdIds`/`droppedIds` so the BFS closure skips them, and `promotedRlsPolicyIds` dedupes across both paths.
- Bonus: the BFS path now also records promotions, so a policy replaced via the pre-existing dependency-closure walk no longer keeps a redundant `ALTER POLICY` — that redundancy existed before this PR.

## One observation (non-blocking)

The create-phase ordering of `CREATE POLICY` **after** `ALTER COLUMN TYPE` has no explicit graph edge: `CreateRlsPolicy.requires` carries table/schema/role ids, none of which have a create-phase producer here, so the correct order comes from logical-sort object-type grouping (tables before policies). That matches how #272/#273 replacements already behave and the inline snapshot pins it — just worth knowing if logical-sort ordering is ever revisited.

Conventions all check out: changeset (patch), conventional commit titles, RED/GREEN commit split per the Test-Driven Fixes policy, `roundtripFidelityTest` + inline snapshot for integration coverage, unit test asserting the design boundary (generic invalidating change, not a concrete subtype), and no SQL parsing — stays strictly within pg_catalog/pg_depend.

---

xmh1011:
@avallete This PR is approved now, and I will leave the cosmetic mock-helper nit unchanged to avoid another push and another review/check reset.

Could you please approve/run the fork CI when you get a chance? Once the required checks are green, I believe this should be ready to merge.

xmh1011:
A bit more context on why I am following up on this one: we depend on pg-toolbelt downstream, so these pg-delta dependency-ordering edge cases affect our migration workflow directly. I have been trying to report the cases with concrete repros and upstream small, reviewable fixes rather than carrying a private patch.

If this PR looks good after the fork CI is approved and passes, could you also include it in the next pg-toolbelt release when that fits your normal release process? Having an upstream release that contains this fix would let downstream users pick it up cleanly without pinning to a fork.

PR #277, opened 2026-06-08
feat(pg-delta): add execution-aware migration units

jgoux:
## Summary

Adds v2 execution-aware migration plans for `@supabase/pg-delta`.

- introduces ordered migration units with explicit transaction modes
- models enum value visibility with structured execution effects
- renders transaction-aware single SQL scripts and numbered migration files
- applies units explicitly with per-unit transaction handling
- preserves legacy flattened plans through centralized normalization
- adds a major changeset for the public API shift

Related: CLI-1596

## Why

PostgreSQL has migration statements whose effects or execution requirements cannot be represented safely as one flat SQL list. In particular, enum values added with `ALTER TYPE ... ADD VALUE` are not usable by later statements until commit, and commands like `CREATE INDEX CONCURRENTLY` cannot run inside a transaction block.

This makes execution context a first-class part of the plan so callers can inspect, render, and apply the migration in the right order and transaction context.

## Future Improvements

Follow-up PRs should consider:

- replacing the current SQL-pattern detection for non-transactional commands with explicit `Change` metadata or parser-backed classification
- exposing an integration annotation hook so custom serializers can provide execution metadata directly
- adding end-to-end coverage for real non-transactional DDL once core diffing can emit statements such as `CREATE INDEX CONCURRENTLY`
- exploring partial-apply resume semantics for multi-unit plans

## Validation

- `cd packages/pg-delta && bun test src/core/plan/execution.test.ts src/core/plan/render.test.ts src/core/plan/normalize.test.ts`
- `cd packages/pg-delta && bun run check-types`
- `bun run format-and-lint`
- `cd packages/pg-delta && bun run knip`
- `cd packages/pg-delta && PGDELTA_TEST_POSTGRES_VERSIONS=17 bun test tests/integration/type-operations.test.ts --test-name-pattern "add enum value|domain CHECK"`
- `cd packages/pg-delta && PGDELTA_TEST_POSTGRES_VERSIONS=17 bun test tests/integration/apply-plan.test.ts`


packages/pg-delta/src/core/plan/execution.ts
```diff
@@ -0,0 +1,569 @@
...
+          name: change.enum.name,
+          stableId: change.enum.stableId,
+        },
+        label: change.newValue,
+      },
+    ],
+  };
+}
+
+function enumConsumerAdapter({
```
avallete: **Correctness gap — incomplete consumer enumeration (needs to be addressed).**

This recognizes a "use" of a newly added enum value only for `AlterTableAlterColumnSetDefault`, `AlterTableAddColumn` (default only — not the generation expression), `CreateTable`, and `AlterTableAddConstraint`. A new value referenced anywhere else in the same transaction will get **no boundary** and re-trigger `55P04`, e.g.:

- view / matview definitions (`... WHERE status = 'store'::user_role`)
- RLS policy `USING` / `WITH CHECK`
- index predicates (`WHERE status = 'store'`)
- partition bounds (`FOR VALUES IN ('store')`)
- generated columns, domain CHECKs

Enumerating every reference site is exactly the fragility we flagged in CLI-1596 — miss one and the bug comes back silently. We need to address this rather than ship the partial matcher as-is.

**Proposed solution (also resolves the substring-matching concern below):** keep the `ExecutionEffect` producer metadata, but make the *boundary decision* conservative and complete instead of precise. When the plan contains any `AlterEnumAddValue`, group all the `ADD VALUE` producers into the leading unit and force a single boundary after them, so every subsequent statement runs in a later (post-commit) unit. This is provably correct regardless of how the value is referenced, removes the per-change-type enumeration and the `expression.includes(...)` text matching entirely, and only costs an extra commit in the rare enum-extend case. Precise matching could later be layered back purely as an *optimization* to merge provably-unrelated statements into the producer unit — but correctness shouldn't depend on it.

---

packages/pg-delta/src/core/plan/execution.ts
```diff
@@ -0,0 +1,569 @@
...
+): ExecutionEffect[] {
+  if (!expression || column.custom_type_category !== "E") return [];
+  if (!column.custom_type_schema || !column.custom_type_name) return [];
+
+  const typeStableId = `type:${column.custom_type_schema}.${column.custom_type_name}`;
+  return addedEnumValues
+    .filter(
+      (effect) =>
+        effect.enumType.stableId === typeStableId &&
+        expression.includes(effect.quotedLabel),
```
avallete: Deciding boundaries by substring-matching the rendered expression text (`expression.includes(effect.quotedLabel)`) is brittle: it's a string scan of `pg_get_expr`-derived SQL living in the plan path, it can false-positive on unrelated text literals (safe — over-splits) and, worse, false-negative whenever the reference site isn't one of the enumerated change types (see the boundary comment on `enumConsumerAdapter`). Prefer the conservative boundary-after-producers approach proposed there, which removes this matching from the correctness path. If we keep effect annotations at all, derive them from the catalog/`pg_depend`, not from scanning serialized SQL.

---

packages/pg-delta/src/core/plan/execution.ts
```diff
@@ -0,0 +1,569 @@
...
+
+function enumEffectsForColumnExpression(
+  column: ColumnProps,
+  expression: string | null,
+  addedEnumValues: AddedEnumValue[],
+): ExecutionEffect[] {
+  if (!expression || column.custom_type_category !== "E") return [];
+  if (!column.custom_type_schema || !column.custom_type_name) return [];
+
+  const typeStableId = `type:${column.custom_type_schema}.${column.custom_type_name}`;
```
avallete: Stable IDs must be built via the `stableId.*` helpers, never inlined as a template literal — `packages/pg-delta/CLAUDE.md` calls this out explicitly ("never inline the prefix as a template literal … was caught in review on this exact pattern"). Same issue at line 314. Use the helper (or the enum model's `stableId` getter) so the format can't drift.

---

packages/pg-delta/src/core/plan/create.ts
avallete, line 1: Two unrelated regressions slipped into this file:

**1. `extractRetries` removed.** `resolveCatalog` now calls `extractCatalog(resolved.pool)` (dropping `{ extractRetries: options.extractRetries }`), and the `extractRetries` field was removed from `CreatePlanOptions` in `types.ts`. That option retries catalog extraction on transient `pg_get_*def()` NULL races during concurrent DDL — unrelated to execution units. Please restore both.

**2. `resolveIntegration` replaced by an inlined reimplementation.** Why? It leaves `resolveIntegration` referenced nowhere (dead export → `knip` failure), and it's the likely cause of the `{ "*/schema": "public" }` → `{ schema: "public" }` change in `apply-plan.test.ts`. If the inlined path no longer supports the `*/…` wildcard filter patterns, that's a silent filter regression masked by editing the test. Please revert to `resolveIntegration` unless there's a concrete reason — and if so, preserve the wildcard-filter behavior and its coverage.

---

packages/pg-delta/tests/integration/apply-plan.test.ts
avallete, line 1: The filter here changed from `{ "*/schema": "public" }` to `{ schema: "public" }`, and the two inline snapshots in this file were downgraded to `toContain(...)` checks. `*/schema` is a real, tested wildcard in the filter DSL (`filter/dsl.test.ts`). If it stopped working after the `resolveIntegration` inlining in `create.ts`, this is a hidden filter regression — please restore the original filter + the exact inline snapshots, or explain why the DSL syntax had to change. Weakening assertions to keep changed output green hides exactly the kind of regression these tests exist to catch.

---

packages/pg-delta/tests/integration/type-operations.test.ts
avallete, line 1: This deletes two existing, unrelated integration tests — "domain CHECK function dependencies are ordered before domains" and "domain CHECK dependency coexists with function using the domain type" — and they aren't moved anywhere. They validate domain→function ordering (via `plan.statements`, which still exists) and have nothing to do with execution units. Please don't drop unrelated passing coverage; restore them, adapting to `flattenPlanStatements`/`plan.statements` if the access pattern changed.

(The two new enum tests added here are great — applying against a live DB and re-diffing to zero is exactly the right end-to-end proof.)

---

avallete (commented):
Thanks for taking this on — the core shape is good: real end-to-end tests that apply against a live DB and re-diff to zero, the single `PoolClient` with session GUCs applied once then per-unit `BEGIN/COMMIT` (correctly preserves `check_function_bodies`), and not rolling back already-committed units. The **breaking major-version direction is fine** (tracked in CLI-1596).

A few things to resolve before merge — details inline:

**Correctness**
- The enum-consumer detection only covers 4 change types and decides boundaries by substring-matching rendered SQL, so a new enum value used by a view / RLS policy / index predicate / partition bound / generated column gets no boundary and re-triggers `55P04`. Proposed a conservative boundary-after-producers approach inline that's complete and drops the fragile text matching.

**Collateral changes that shouldn't be in this PR**
- `extractRetries` was removed from `createPlan`/`CreatePlanOptions` (unrelated catalog-extraction retry feature).
- `resolveIntegration` was replaced by an inlined reimplementation (now a dead export; likely the cause of the filter-DSL change below).

**Tests**
- Two unrelated domain-ordering integration tests were deleted rather than ported.
- A filter test was switched from the `*/schema` wildcard to `schema`, and two inline snapshots were weakened to `toContain` — please confirm the wildcard filter didn't regress and restore the precise assertions.

**Nit:** stable IDs are built as inline `type:${…}` template literals — use the `stableId.*` helpers per CLAUDE.md.

Also worth linking this PR to CLI-1596. Happy to re-review once these are addressed.

---

packages/pg-delta/src/core/plan/execution.ts
jgoux: Addressed. I removed the enum-consumer enumeration and no longer rely on detecting every possible reference site. The planner now treats enum committed effects conservatively: once a transactional unit has produced enum values, the next non-producer statement is forced into a later unit. That means views, RLS policies, predicates, generated columns, partition-related statements, and other opaque consumers all run after the enum producer transaction commits. I added coverage for an opaque view-like consumer to lock this behavior in.

packages/pg-delta/src/core/plan/execution.ts
jgoux: Addressed. The expression substring matching path has been removed from the correctness model. Enum value effects are still emitted on producer statements, but unit splitting is now conservative after enum producers rather than based on scanning rendered expressions.

packages/pg-delta/src/core/plan/execution.ts
jgoux: Addressed. The inline type stable-id construction was removed with the consumer scanner. The remaining enum effect metadata uses the enum model\u2019s stableId getter, so production code is no longer building type stable IDs with ad hoc template literals.

packages/pg-delta/src/core/plan/create.ts
jgoux: Addressed. The branch is rebased on current main and now preserves both main-branch behaviors: createPlan forwards extractRetries into extractCatalog, and buildPlanForCatalogs uses resolveIntegration instead of an inline reimplementation. I also ran knip; it no longer reports resolveIntegration as a dead export.

packages/pg-delta/tests/integration/apply-plan.test.ts
jgoux: Addressed. The filter test is back to the { "*/schema": "public" } wildcard form and the exact inline snapshots are restored. The targeted apply-plan integration test passes against PG17.

packages/pg-delta/tests/integration/type-operations.test.ts
jgoux: Addressed. Both domain-ordering integration tests are restored unchanged in purpose, still checking plan.statements ordering and dependency extraction. I included them in the targeted PG17 validation with the domain CHECK test-name pattern.

packages/pg-delta/src/core/plan/execution.ts
```diff
@@ -0,0 +1,342 @@
...
+  return result;
+}
+
+function requiresNoTransaction(sql: string): boolean {
+  return splitStatements(stripSqlComments(sql)).some((statement) =>
+    NON_TRANSACTIONAL_PATTERNS.some((pattern) => pattern.test(statement)),
+  );
+}
+
+const NON_TRANSACTIONAL_PATTERNS = [
```
avallete, line 304: Non-blocking: this `transactionMode: "none"` / `NON_TRANSACTIONAL_PATTERNS` machinery is still dead — pg-delta's core can't emit `CREATE INDEX CONCURRENTLY` / `REINDEX … CONCURRENTLY` / etc. today, so `requiresNoTransaction` never matches real output and there's no end-to-end coverage for it. That's fine as scaffolding (the PR body lists it as future work), just flagging that it ships untested against actual emitted SQL. When core can emit such statements, the follow-up should classify them via `Change` metadata rather than regex-matching rendered SQL.

---

packages/pg-delta/src/core/plan/execution.ts
```diff
@@ -0,0 +1,342 @@
...
+    producesCommittedEffects: uniqueEffects(
+      annotation.producesCommittedEffects,
+    ),
+  };
+}
+
+function emptyAnnotation(): StatementAnnotation {
+  return {
+    requiresCommittedEffects: [],
+    producesImmediateEffects: [],
```
avallete: Non-blocking: `producesImmediateEffects` is now vestigial — no adapter ever populates it (the producer adapter only sets `producesCommittedEffects`), so it's always `[]`. Either wire it up to a real use or drop it from `PlannedStatement`/the schema to avoid carrying a field that always reads empty.

---

packages/pg-delta/src/core/plan/execution.ts
```diff
@@ -0,0 +1,342 @@
...
+  const sessionStatements = buildSessionStatements(changes, options);
+  const plannedStatements = changes.map((change, index) =>
+    planStatement(change, index, options.integration),
+  );
+  const units = buildMigrationUnits(plannedStatements);
+
+  return {
+    units,
+    sessionStatements,
+    statements: [
```
avallete, line 63: Non-blocking: the plan now persists three overlapping views of the same SQL — `units`, the flattened `statements` (which also re-includes `sessionStatements`), and `sessionStatements`. Storing all three in the serialized plan bloats it and risks divergence on round-trip (e.g. if a consumer mutates one but not the others). Consider treating `units` (+ `sessionStatements`) as the single source of truth and deriving `statements` on demand via `flattenPlanStatements`/`renderPlanSql` rather than persisting it.

---

.changeset/execution-units-major.md
avallete, line 1: Non-blocking: this is user-facing and a `major`, so it's worth a sentence on the behavioral consequences, not just the new capability — specifically that multi-unit plans are **no longer fully atomic** (a forced commit boundary means earlier units commit before later ones), and that on partial failure already-committed units are not rolled back (and an added enum value can't be dropped). That sets the right expectation for anyone consuming the new plan shape.

---

avallete (commented):
Thanks for the quick turnaround — this revision addresses everything from the previous round:

- **Enum boundary detection** is now the conservative, provably-correct model: producers group together and any subsequent non-`ADD VALUE` statement is pushed past a commit boundary, regardless of how it references the value. The brittle per-change-type enumeration + `expression.includes(...)` substring matching are gone, which closes the false-negative class (views / RLS / index predicates / generated columns / partition bounds). The new "splits after enum values before opaque later statements" unit test captures exactly that guarantee. 👍
- `extractRetries` and `resolveIntegration` restored (and `knip` added to validation).
- The two domain-ordering integration tests are back.
- The `*/schema` wildcard filter and the exact inline snapshots are restored.
- Inline `type:${…}` stable-id literals replaced with `change.enum.stableId`.

I traced the rewritten `buildMigrationUnits` through the grouped, single-consumer, and interleaved cases and it holds up.

A few **minor, non-blocking** notes inline (speculative non-transactional regex machinery is still dead/untested; `producesImmediateEffects` is vestigial; `units`+`statements`+`sessionStatements` triple-storage; changeset could mention the atomicity/partial-apply consequences). None of these need to gate merge.

Nice work — core design landed in the right place.

---

packages/pg-delta/src/core/plan/execution.ts
jgoux: Keeping this as scaffolding for this PR rather than changing it: the review is right that core does not currently emit these statements, and the PR body keeps parser/Change-metadata classification as future work. The current behavior is covered at the unit/render/apply-unit boundary but not claimed as end-to-end core output. I am leaving the implementation in place because the public transactionMode: "none" path needs a concrete planner route, and the follow-up should replace regex detection once core can emit those changes.

packages/pg-delta/src/core/plan/execution.ts
jgoux: Addressed. Removed producesImmediateEffects from PlannedStatement, PlanSchema, legacy normalization, planner annotations, and tests. The only remaining effect channel is committed effects, which is the one used by enum visibility.

packages/pg-delta/src/core/plan/execution.ts
jgoux: Keeping statements for this PR as deprecated compatibility rather than removing it now. Existing tests and likely users still consume plan.statements; removing it would broaden this PR beyond execution units. units remains the primary v2 surface and render/apply paths use units. I agree the eventual cleanup is to derive flattened statements on demand in a follow-up once consumers have migrated.

.changeset/execution-units-major.md
jgoux: Addressed. Expanded the major changeset to call out that multi-unit plans are not fully atomic across the whole migration, that later failures do not roll back already committed units, and that enum values are not automatically removable after commit.

packages/pg-delta/src/core/plan/apply.ts
```diff
@@ -152,13 +170,12 @@ export async function applyPlan(
     }
 
     // Count only actual changes, excluding session configuration statements
-    const changeStatements = statements.filter(
-      (stmt) => !isSessionStatement(stmt),
-    );
+    const changeStatements = units.flatMap((unit) => unit.statements);
```
avallete, line 173: **issue**

I think there is an issue here, each statement should still pass the `isSessionStatements` ?

avallete (commented):
Since it is a major and large breaking migration, I think we should document this decision as an ADR / update the CLAUDE.md/AGENT.md intruction so  when we add / update statements this is taken into account and reviewed.

Also the changeset should contain clear BREAKING CHANGE intstruction with migration examples for api consumer to ease the transition.

Another thing I want to make sure of is what happen for statements that cannot be within an actual transaction ? How are they detected / treated ?

packages/pg-delta/src/core/plan/execution.ts
```diff
@@ -0,0 +1,333 @@
...
+  );
+}
+
+const NON_TRANSACTIONAL_PATTERNS = [
+  /^\s*CREATE\s+(?:UNIQUE\s+)?INDEX\s+CONCURRENTLY\b/i,
+  /^\s*DROP\s+INDEX\s+CONCURRENTLY\b/i,
+  /^\s*REINDEX\b[\s\S]*\bCONCURRENTLY\b/i,
+  /^\s*REFRESH\s+MATERIALIZED\s+VIEW\s+CONCURRENTLY\b/i,
+  /^\s*ALTER\s+TABLE\b[\s\S]*\bDETACH\s+PARTITION\b[\s\S]*\bCONCURRENTLY\b/i,
+];
```
avallete, line 310: **The non-transactional classifier is live on every emitted statement — and has both a false-positive and a false-negative we can now name concretely.**

This was discussed last round as "dead scaffolding", but note it isn't inert: `requiresNoTransaction` runs over every statement in every plan, including `CREATE FUNCTION` bodies from `pg_get_functiondef`.

**False positive (verified by repro):** `splitStatements` splits on `;` with no awareness of dollar-quoting or string literals, so a perfectly transactional function whose body contains multi-statement dynamic SQL gets misclassified:

```sql
CREATE FUNCTION public.maint() RETURNS void LANGUAGE plpgsql AS $$
BEGIN
  EXECUTE 'SET lock_timeout = ''1s''; CREATE INDEX CONCURRENTLY idx ON public.t (c)';
END;
$$
```

→ `requiresNoTransaction(...)` returns `true`, and the whole `CREATE FUNCTION` is pulled out of its transactional unit into a standalone autocommit unit. Maintenance helpers that build concurrent-index DDL strings are a real pattern, and the result is a silent loss of atomicity plus extra commit boundaries.

**False negative:** Codex's point below about `CREATE SUBSCRIPTION` is correct — when `replication_slot_created` is true, `CreateSubscription.serialize()` emits a plain `CREATE SUBSCRIPTION` (default `connect = true`), which PostgreSQL rejects inside a transaction block, and `applyUnit` now always wraps transactional units in explicit `BEGIN`/`COMMIT`. Unlike `CREATE INDEX CONCURRENTLY`, this is SQL that pg-delta itself emits today, so it deserves handling in this PR rather than the follow-up.

**Suggestion that fixes both:** the adapter mechanism already receives `{ change, sql }` — classify by `change instanceof CreateSubscription` (plus the slot/connect condition) instead of regex-matching `sql`, and scope the regex fallback so it never inspects statements containing `$$`/dollar-quote tags (or drop the regex path entirely until core can emit concurrent DDL). That turns the planned "Change metadata classification" follow-up into a one-class change now, with the subscription case as its first real test.

---

packages/pg-delta/src/core/plan/normalize.ts
```diff
@@ -0,0 +1,56 @@
...
+  const schemaStatements = statements.filter(
+    (statement) => !isSessionStatement(statement),
+  );
+  if (schemaStatements.length === 0) return [];
+
+  return [
+    {
+      id: "unit_001",
+      name: "schema_changes",
+      transactionMode: "transactional",
```
avallete, line 46: **Legacy-plan semantics: the single-transaction wrap is faithful for pg-delta-generated plans, with one narrow exception (Codex's point holds for that case).**

I checked what `main`'s `applyPlan` actually did: it joined all statements and ran them as one multi-statement `pool.query(script)` — the simple query protocol wraps that in an **implicit transaction**. So wrapping legacy `statements` in a single transactional unit here preserves the old semantics exactly for multi-statement plans — no regression there. 👍

The exception is a legacy plan whose only change statement is non-transactional (e.g. a hand-edited plan with one `CREATE INDEX CONCURRENTLY`): the old path ran a single statement under autocommit and succeeded; the new path adds `BEGIN`/`COMMIT` and fails. pg-delta never emitted such statements itself, so this is an edge case — but the fix is cheap and already in this PR's vocabulary: run each legacy statement through the same non-transactional classification used by `buildExecutionPlan` and emit `transactionMode: "none"` units for matches, instead of unconditionally `"transactional"`.

---

packages/pg-delta/src/core/plan/render.ts
```diff
@@ -0,0 +1,145 @@
...
+      includeFileHeader: true,
+    }),
+    unit,
+  }));
+}
+
+export function flattenPlanStatements(plan: Plan): string[] {
+  const units = planUnits(plan);
+  if (units.length === 0) return plan.statements;
+  return units.flatMap((unit) => unit.statements.map((stmt) => stmt.sql));
```
avallete, line 57: **Confirmed: `flattenPlanStatements` drops `sessionStatements` for unit-based plans (Codex is right).**

For a v2 plan this returns only the unit statements, while the deprecated `plan.statements` field it's meant to replace includes `SET ROLE` / `SET check_function_bodies = false` (see `buildExecutionPlan`, which builds `statements` as `[...sessionStatements, ...unitSql]`). A caller migrating from `plan.statements` to this helper silently loses the role scoping. The empty-units fallback returning `plan.statements` (session statements included) makes the helper inconsistent even with itself.

One-line fix:

```ts
const normalized = normalizePlan(plan);
if (normalized.units.length === 0) return normalized.statements;
return [
  ...(normalized.sessionStatements ?? []),
  ...normalized.units.flatMap((unit) => unit.statements.map((stmt) => stmt.sql)),
];
```

Worth a small unit test in `render.test.ts` asserting parity with the deprecated `plan.statements` for a role-scoped plan.

---

packages/pg-delta/src/core/plan/execution.ts
```diff
@@ -0,0 +1,333 @@
...
+}
+
+interface ExecutionPlan {
+  units: MigrationUnit[];
+  sessionStatements: string[];
+  statements: string[];
+}
+
+type StatementAnnotation = {
+  requiresCommittedEffects: ExecutionEffect[];
```
avallete, line 26: Nit, same class as the `producesImmediateEffects` removal last round: `requiresCommittedEffects` is never populated by production code — no adapter sets it (only the enum *producer* side sets `producesCommittedEffects`), and `requiredBoundaryReason` doesn't read it. Every real plan serializes it as `[]` on every statement; only the test fixture in `render.test.ts` fills it in by hand. Either populate it where the boundary is created (annotating the first statement of an `after_enum_values` unit with the effects it waited on would make plans self-describing) or drop it from `PlannedStatement`/`PlanSchema`.

While in here: the `_ctx: DiffContext` parameter of `buildExecutionPlan` is also unused (leftover from the removed consumer scanner) — the unit tests construct mock catalogs that nothing reads. Dropping it simplifies both the API and the tests.

---

packages/pg-delta/src/core/plan/apply.ts
avallete, line 173: Re the open question on this line ("should each statement still pass `isSessionStatement`?") — checked, and it's correct as written, no filter needed. Session statements can never appear inside `units`: `buildExecutionPlan` routes `SET ROLE` / `SET check_function_bodies` exclusively into `sessionStatements` (built in `buildSessionStatements`, never planned as unit statements), and for legacy v1 plans `legacyUnits` in `normalize.ts` explicitly filters `isSessionStatement` matches out of the unit it builds. So `units.flatMap(u => u.statements)` is session-free by construction — the old `.filter(stmt => !isSessionStatement(stmt))` was only needed because the flat `plan.statements` array mixed both. The comment above the line still holds.

---

avallete (commented):
Full re-review of the current head (`18756b9`, the merge of `main`), including verification of the three Codex findings and answers to the open questions from the previous review. CI is green across the board (all 45 integration shards on PG 15/17/18, unit tests, types, lint, knip, CodeQL), and I re-ran the plan-module unit tests, `check-types`, `format-and-lint`, and `knip` locally — all pass. The merge with `main` (#273/#275) is clean; those touch the sort/diff layers and don't interact with unit grouping.

## Answering the open questions

**"What happens to statements that cannot run inside a transaction?"** They are detected by regex (`NON_TRANSACTIONAL_PATTERNS` in `execution.ts`) over each statement's rendered SQL — after comment-stripping and a naive `;` split — and placed in their own `transactionMode: "none"` unit. `applyUnit` executes such units statement-by-statement without `BEGIN`/`COMMIT`, and the renderer emits them without transaction wrappers. Two problems with the detection as shipped, both detailed inline: a **false positive** (dollar-quoted function bodies containing multi-statement dynamic SQL get misclassified and lose atomicity — repro included) and a **false negative** that Codex correctly spotted (`CREATE SUBSCRIPTION` with `connect = true` is emitted by pg-delta today, is rejected by PostgreSQL inside a transaction block, and now always gets wrapped). The subscription case is the one I'd fix in this PR — it's pg-delta's own output, and an `instanceof CreateSubscription` check in the adapter is the natural first instance of the Change-metadata classification we already agreed should replace the regexes.

**Codex findings, triaged:** all three are real. The `CREATE SUBSCRIPTION` one is the substantive one (above). The legacy-plan one is narrower than stated — I verified `main`'s applier ran the whole script as one multi-statement query, i.e. an *implicit transaction*, so the single-transaction wrap is faithful except for single-statement legacy plans with non-transactional SQL (details inline, cheap fix available). The `flattenPlanStatements` one is a genuine API inconsistency — it drops `sessionStatements` that the deprecated `plan.statements` field includes — with a one-line fix (inline).

**ADR / agent-docs:** agreed. Concretely: a short section in `packages/pg-delta/CLAUDE.md` stating the execution model invariants — (1) any new change class whose SQL cannot run in a transaction block must be classified in `execution.ts` via Change metadata, not regex; (2) any new "effect not visible until commit" semantics must go through `ExecutionEffect` producers + the conservative boundary rule; (3) `units` is the execution source of truth and `plan.statements` is render-only legacy. That turns this review's lessons into a checked-in rule for future statement work.

**Changeset migration guidance:** agreed — the breaking-change text now states the atomicity consequences but gives consumers no before/after. A short snippet (`plan.statements.join(";\n")` → `renderPlanSql(plan)` / `renderPlanFiles(plan)`, and `applyPlan` handling `failedUnitId`/`completedUnitIds`) would make the major bump self-service.

## Remaining nits (inline)

`requiresCommittedEffects` is vestigial in production (same class as the removed `producesImmediateEffects`), and `buildExecutionPlan`'s `_ctx` parameter is dead. Also answered the open `apply.ts` question inline: the statement count needs no `isSessionStatement` filter — units are session-free by construction.

## Verdict

Core design is in good shape and everything from rounds 1–2 held up under re-verification. Before merge I'd resolve: (1) `CREATE SUBSCRIPTION` classification, (2) the `flattenPlanStatements` session-statement fix, and (3) the changeset migration example. The dollar-quote false positive and legacy non-transactional split can ride along now or land in the agreed follow-up — but if the regex path stays, it should at least skip statements containing dollar-quoted bodies.

---

PR #278, opened 2026-06-08
fix(pg-delta): order dependent view drops before column type rewrites

avallete:
## Summary

Fixes ordering of schema changes when a column type is altered in place and dependent views exist. Views must be dropped before the column type rewrite and recreated after, since the rewrite invalidates the view's definition.

## Changes

- **Added `invalidates` property to `BaseChange`**: A new getter that returns stable identifiers of objects that are mutated in place. Unlike `drops`, the object keeps its identity but its dependents must be torn down and rebuilt around the mutation.

- **Implemented `invalidates` in `AlterTableAlterColumnType`**: Reports the column being altered as invalidated, since `ALTER COLUMN ... TYPE` rewrites the column in place and breaks dependent views.

- **Updated graph builder to handle invalidations**: In the drop phase, invalidated ids are treated like dropped ids for ordering purposes — dependents are ordered before the mutation, ensuring views are dropped before the column type change.

- **Added test coverage**: 
  - Unit test in `sort-changes.test.ts` verifying correct ordering: `DropView` → `AlterTableAlterColumnType` → `CreateView`
  - Integration test in `alter-table-operations.test.ts` validating the full roundtrip with actual database operations

- **Added changeset** for patch release

## Implementation Details

The `invalidates` mechanism is distinct from `drops`:
- `drops`: Object is removed entirely; phase assignment, filtering, and serialization all change
- `invalidates`: Object is mutated in place; only ordering is affected via the dependency graph

This allows the sorter to use `pg_depend` edges to order dependent teardowns before the mutation, without changing how the mutation itself is classified or serialized.

https://claude.ai/code/session_01Fnhs5Jt3LhiLQzF1c8JguH


avallete:
Closing in favor of: https://github.com/supabase/pg-toolbelt/pull/273

PR #279, opened 2026-06-09
fix(pg-delta): normalize function body line endings

### LINKED ISSUE #242: pg-delta treats CRLF vs LF inside function bodies as a schema diff
## Affected package

`pg-delta`

## Version / commit / branch tested

Observed through Supabase CLI `2.95.2`, which uses `pg-delta` for declarative schema diff/sync.

If helpful, the generated pgdelta temp catalogs were produced against PostgreSQL `17.6`.

## Environment

- OS: Windows
- Shell/runtime: PowerShell
- Database: Supabase local / Supabase declarative schema workflow
- PostgreSQL: `17.6`

## Summary

`pg-delta` appears to compare PostgreSQL function bodies with line endings preserved. If a function was originally created from SQL using CRLF line endings, and the declarative schema contains the same function body with LF line endings, `pg-delta` reports a function diff and emits `CREATE OR REPLACE FUNCTION`, even though the function is semantically identical.

This causes a no-op declarative sync to generate a migration containing only function replacements.

### Source state SQL

Create a function from a SQL file using CRLF line endings:

```sql
CREATE OR REPLACE FUNCTION public.set_updated_at()
RETURNS trigger
LANGUAGE plpgsql
AS $$
begin
  new.updated_at = now();
  return new;
end;
$$;
```

The important part is that the file contains CRLF line endings, so the stored function source includes `\r\n`.

### Target declarative schema

Declare the same function in a file using LF line endings:

```sql
CREATE FUNCTION public.set_updated_at()
  RETURNS TRIGGER
  LANGUAGE plpgsql
  AS $function$
begin
  new.updated_at = now();
  return new;
end;
$function$;
```

### Command

This was observed through Supabase declarative schema sync/diff.

The generated diff produced:

```sql
SET check_function_bodies = false;

CREATE OR REPLACE FUNCTION public.set_updated_at()
 RETURNS trigger
 LANGUAGE plpgsql
AS $function$
begin
  new.updated_at = now();
  return new;
end;
$function$;
```

## Expected result

No diff should be generated, because the function body is semantically identical.

At minimum, CRLF and LF line endings inside function source should be normalized before comparing function definitions / source code.

## Actual result

`pg-delta` detects a difference in the function body and emits `CREATE OR REPLACE FUNCTION`.

The catalog comparison shows the only difference is line endings:

```text
source_code in migration-created DB:
"\r\nbegin\r\n  new.updated_at = now();\r\n  return new;\r\nend;\r\n"

source_code in declarative-created DB:
"\nbegin\n  new.updated_at = now();\n  return new;\nend;\n"
```

After applying the generated replacement migration, a subsequent sync is clean because the function body has been rewritten with LF line endings.

## Why this matters

On Windows, an initial migration may be authored with CRLF line endings while generated declarative schema files use LF. That makes the first declarative sync after initialization produce an unexpected migration even though there are no real schema changes.

In a larger schema, this can re-emit every function created from the CRLF migration, creating noisy migrations and making it harder to distinguish real schema drift from formatting drift.

## Smallest failing scenario

One PL/pgSQL function is enough to reproduce:

1. Apply migration SQL with CRLF line endings.
2. Compare against declarative schema SQL with LF line endings.
3. Observe `CREATE OR REPLACE FUNCTION` emitted.
4. Normalize the original migration to LF and repeat.
5. Observe no diff.

## Supabase-specific details

This was observed in a Supabase project using declarative schema files under:

```text
supabase/database/schemas/public/functions/
```

The generated no-op sync migration contained only `CREATE OR REPLACE FUNCTION` statements.

I do not believe this depends on a Supabase-managed object, extension, policy, or role. The issue appears to be limited to function body source comparison.

pjpjq:
## Summary

- Normalize CRLF/LF differences in `source_code`, `sql_body`, and `definition` when building the procedure stable snapshot.
- Reuse the same line-ending normalization for body-only non-alterable comparisons to avoid false `CREATE OR REPLACE FUNCTION` diffs.
- Add regression coverage showing line-ending-only differences do not produce a diff while owner-only changes are still detected.

Fixes #242.
Replaces #276.

## Tests

- `bun test packages/pg-delta/src/core/objects/procedure/procedure.diff.test.ts`
- `cd packages/pg-delta && bun run check-types`
- `cd packages/pg-delta && bun run format-and-lint`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/packages/pg-delta/src/core/objects/procedure/procedure.diff.test.ts
+++ b/packages/pg-delta/src/core/objects/procedure/procedure.diff.test.ts
@@ -162,6 +162,63 @@ describe.concurrent("procedure.diff", () => {
     expect((changes[0] as CreateProcedure).orReplace).toBe(true);
   });
 
+  test("does not replace a function when body line endings differ only by CRLF vs LF", () => {
+    const bodyLf = "\nbegin\n  new.updated_at = now();\n  return new;\nend;\n";
+    const bodyCrlf = bodyLf.replace(/\n/g, "\r\n");
+    const definitionLf = `CREATE FUNCTION public.fn1() RETURNS trigger LANGUAGE plpgsql AS $function$${bodyLf}$function$`;
+    const definitionCrlf = definitionLf.replace(/\n/g, "\r\n");
+    const main = new Procedure({
+      ...base,
+      return_type: "trigger",
+      language: "plpgsql",
+      source_code: bodyCrlf,
+      definition: definitionCrlf,
+    });
+    const branch = new Procedure({
+      ...base,
+      return_type: "trigger",
+      language: "plpgsql",
+      source_code: bodyLf,
+      definition: definitionLf,
+    });
+    const changes = diffProcedures(
+      testContext,
+      { [main.stableId]: main },
+      { [branch.stableId]: branch },
+    );
+    expect(changes).toEqual([]);
+  });
+
+  test("body line ending differences do not mask an owner-only function diff", () => {
+    const bodyLf = "\nbegin\n  new.updated_at = now();\n  return new;\nend;\n";
+    const bodyCrlf = bodyLf.replace(/\n/g, "\r\n");
+    const definitionLf = `CREATE FUNCTION public.fn1() RETURNS trigger LANGUAGE plpgsql AS $function$${bodyLf}$function$`;
+    const definitionCrlf = definitionLf.replace(/\n/g, "\r\n");
+    const main = new Procedure({
+      ...base,
+      return_type: "trigger",
+      language: "plpgsql",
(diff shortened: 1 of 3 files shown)


avallete:
I think the issue with such approach is that the function itself might contain an actual `"\r\n"` in it, then I wonder how the normalization would go 🤔 

I would rather not do any treatment to functions body and so on, those should be black box user input that are never changed and compared as is. I think the underneath issue of the original issue is that there was discrepancies between the windows/unix(that postgres is) new lines ending. And in this case, the solution is to set the terminal/filesystem to use LF as line ending rather than normalize the diff in pg-delta.

PR #287, opened 2026-06-11
fix(pg-topo): order alter table expression functions

### LINKED ISSUE #281: pg-topo: ALTER TABLE expression subcommands miss function dependencies
## Summary

While deep-testing current `main` (`9f01826b00b64e7f9628844ecb6f69a07f53c1f8`), I found a `pg-topo` static dependency extraction gap for `ALTER TABLE` subcommands whose expressions call user-defined functions.

This is one issue class, so I am grouping the two repros together instead of opening separate issues:

1. `ALTER TABLE ... ADD COLUMN ... DEFAULT app.fn()`
2. `ALTER TABLE ... ADD CONSTRAINT ... CHECK (app.fn(col))`

Both statements are sorted before the function they call, then fail at runtime with `42883`.

## Duplicate / related issue check

I checked existing issues and PRs before opening this:

- #115 is a shared sorting abstraction proposal, not a concrete `ALTER_TABLE` extraction bug.
- #135 / #148 covered `pg-delta` domain `CHECK` function dependencies via `pg_depend`; this repro is in `@supabase/pg-topo` static SQL extraction and involves `ALTER TABLE` subcommands.
- #277 is `pg-delta` execution-unit work and does not touch `pg-topo` expression extraction.
- Searches for `pg-topo ALTER TABLE ADD COLUMN`, `ALTER TABLE ADD CHECK`, `CHECK function dependency`, `pg-topo check constraint`, `index expression dependency`, and `publication row filter dependency` did not show an existing issue/PR covering this exact `pg-topo` case.

## Repro A: ADD COLUMN DEFAULT calling a function

Input statements, intentionally shuffled:

```sql
alter table app.items add column slug text default app.default_slug();

create function app.default_slug()
returns text
language sql
immutable
as $$ select 'untitled' $$;

create table app.items(id int primary key, name text);

create schema app;
```

Actual `pg-topo` order:

```text
create schema app;
create table app.items(id int primary key, name text);
alter table app.items add column slug text default app.default_slug();
create function app.default_slug() returns text language sql immutable as $$ select 'untitled' $$;
```

Runtime validation diagnostic:

```text
code: RUNTIME_EXECUTION_ERROR
message: function app.default_slug() does not exist
sqlstate: 42883
missingObjectKey: function:app:default_slug:()
producerStatementId: <input:3>:0
```

## Repro B: ADD CHECK calling a function

Input statements, intentionally shuffled:

```sql
alter table app.items
  add constraint items_name_check check (app.is_valid_name(name));

create function app.is_valid_name(value text)
returns boolean
language sql
immutable
as $$ select length(value) > 0 $$;

create table app.items(id int primary key, name text);

create schema app;
```

Actual `pg-topo` order:

```text
create schema app;
create table app.items(id int primary key, name text);
alter table app.items add constraint items_name_check check (app.is_valid_name(name));
create function app.is_valid_name(value text) returns boolean language sql immutable as $$ select length(value) > 0 $$;
```

Runtime validation diagnostic:

```text
code: RUNTIME_EXECUTION_ERROR
message: function app.is_valid_name(text) does not exist
sqlstate: 42883
missingObjectKey: function:app:is_valid_name:(text)
producerStatementId: <input:3>:0
```

## Expected behavior

`pg-topo` should extract expression dependencies from `ALTER TABLE` subcommands and add edges from the `ALTER_TABLE` statement to referenced functions/types/tables, so the order becomes:

```text
create schema app;
create table app.items(...);
create function app.default_slug(...);
alter table app.items add column ... default app.default_slug();
```

and similarly for `ADD CONSTRAINT ... CHECK`.

## Root cause

`packages/pg-topo/src/extract/extract-dependencies.ts::extractAlterTableDependencies` currently handles only a narrow subset of `AlterTableCmd` dependency surfaces:

- the target relation itself;
- foreign-key dependencies;
- primary/unique key provider metadata;
- owner changes;
- partition attachment.

It does not recurse into expression-bearing `AlterTableCmd.def` shapes.

The parser exposes these dependencies in the AST:

- `ADD COLUMN ... DEFAULT app.default_slug()`:
  `AlterTableCmd.def.ColumnDef.constraints[].Constraint.raw_expr`
- `ADD CONSTRAINT ... CHECK (app.is_valid_name(name))`:
  `AlterTableCmd.def.Constraint.raw_expr`

`CREATE TABLE` already has similar extraction logic for `ColumnDef.constraints` and table `Constraint.raw_expr`, but the `ALTER_TABLE` path does not mirror it. As a result, `buildGraph` never sees `function:app:default_slug:()` or `function:app:is_valid_name:(text)` as requirements of the `ALTER_TABLE` statement, and the deterministic class ordering places `ALTER_TABLE` before `CREATE_FUNCTION`.

## Suggested fix

Extend `extractAlterTableDependencies` to mirror the relevant `CREATE_TABLE` expression extraction behavior for `AlterTableCmd.def`:

1. When `command.def` is a `ColumnDef`:
   - add the column type dependency via `typeFromTypeNameNode(columnDefinition.typeName)`;
   - inspect `columnDefinition.constraints`;
   - for `Constraint.raw_expr`, call `addExpressionDependencies(raw_expr, requires)`;
   - preserve existing FK / PK / UNIQUE handling where applicable.
2. When `command.def` is a `Constraint`:
   - preserve existing FK / PK / UNIQUE handling;
   - for `constraint.raw_expr`, call `addExpressionDependencies(raw_expr, requires)`.
3. Add focused `pg-topo` tests with runtime validation for both repros above.

## Local validation

I used a local-only probe test on current `main`, with `postgres:17-alpine` temporarily tagged from the already-present `postgres:17.6-alpine` image so the `pg-topo` global setup did not block on a pull.

Commands:

```bash
docker tag postgres:17.6-alpine postgres:17-alpine

./node_modules/.bin/bun run --cwd packages/pg-topo scripts/run-tests.ts \
  --test-name-pattern "orders ALTER TABLE ADD COLUMN default" \
  test/deep-probe.test.ts

./node_modules/.bin/bun run --cwd packages/pg-topo scripts/run-tests.ts \
  --test-name-pattern "orders ALTER TABLE ADD CHECK" \
  test/deep-probe.test.ts
```

Both focused repros fail consistently with the runtime diagnostics shown above.

xmh1011:
## Summary

Closes #281.

This fixes missing pg-topo dependencies for expression-bearing `ALTER TABLE` subcommands. The extractor now reuses the existing expression dependency traversal for:

- `ADD COLUMN ... DEFAULT app.fn(...)`
- `ADD CONSTRAINT ... CHECK (app.fn(...))`
- `ALTER COLUMN ... SET DEFAULT app.fn(...)`
- `ADD COLUMN ... GENERATED ALWAYS AS (app.fn(...)) STORED`
- `ALTER COLUMN ... TYPE ... USING app.fn(...)`

## RED evidence

Before the extractor change, the focused regression sorted `ALTER TABLE` before the referenced function for the cases above. The generated-column and `USING` variants also surfaced runtime execution errors against PostgreSQL.

## Validation

```bash
bun run --cwd packages/pg-topo scripts/run-tests.ts test/alter-table-expression-dependencies.test.ts
```

Result: `5 pass / 0 fail`.

```bash
bun run test:pg-topo
```

Result: `149 pass / 0 fail`.

```bash
bun run format-and-lint
bun run check-types
bun run knip
```

Result: passed. `knip` only reported existing configuration hints.

### DIFF
--- a/packages/pg-topo/src/extract/extract-dependencies.ts
+++ b/packages/pg-topo/src/extract/extract-dependencies.ts
@@ -163,7 +163,6 @@ const addConstraintExpressionDependencies = (
   }
   for (const expressionNode of [
     constraint.raw_expr,
-    constraint.cooked_expr,
     constraint.exclusions,
     constraint.where_clause,
   ]) {
@@ -213,11 +212,35 @@ const extractAlterTableDependencies = (
     if (columnDefinition?.raw_default) {
       addExpressionDependencies(columnDefinition.raw_default, requires);
     }
+    const typeRef = typeFromTypeNameNode(columnDefinition?.typeName);
+    if (typeRef) {
+      requires.push(typeRef);
+    }
     const columnConstraints = Array.isArray(columnDefinition?.constraints)
       ? columnDefinition.constraints
       : [];
     for (const constraintItem of columnConstraints) {
       const columnConstraint = asRecord(asRecord(constraintItem)?.Constraint);
+      if (columnConstraint?.contype === "CONSTR_FOREIGN") {
+        addForeignConstraintDependencies(columnConstraint, requires);
+      }
+      if (
+        relationTableRef &&
+        (columnConstraint?.contype === "CONSTR_PRIMARY" ||
+          columnConstraint?.contype === "CONSTR_UNIQUE")
+      ) {
+        const columnName =
+          typeof columnDefinition?.colname === "string"
+            ? columnDefinition.colname
+            : undefined;
+        const providedKey = keyRefForTableColumns(
+          relationTableRef,
+          constraintKeyColumns(columnConstraint, columnName),
+        );
+        if (providedKey) {
+          provides.push(providedKey);
+        }
+      }
       addConstraintExpressionDependencies(columnConstraint, requires);
     }
 
--- a/packages/pg-topo/src/extract/extract-dependencies.ts
+++ b/packages/pg-topo/src/extract/extract-dependencies.ts
@@ -172,6 +172,7 @@ const extractAlterTableDependencies = (
   for (const commandNode of commands) {
     const command = asRecord(asRecord(commandNode)?.AlterTableCmd);
     const constraint = asRecord(asRecord(command?.def)?.Constraint);
+    const columnDefinition = asRecord(asRecord(command?.def)?.ColumnDef);
     if (constraint?.contype === "CONSTR_FOREIGN") {
       addForeignConstraintDependencies(constraint, requires);
     }
@@ -187,6 +188,26 @@ const extractAlterTableDependencies = (
       if (providedKey) {
         provides.push(providedKey);
       }
+    }
+    if (constraint?.raw_expr) {
+      addExpressionDependencies(constraint.raw_expr, requires);
+    }
+
+    if (columnDefinition?.raw_default) {
+      addExpressionDependencies(columnDefinition.raw_default, requires);
+    }
+    const columnConstraints = Array.isArray(columnDefinition?.constraints)
+      ? columnDefinition.constraints
+      : [];
+    for (const constraintItem of columnConstraints) {
+      const columnConstraint = asRecord(asRecord(constraintItem)?.Constraint);
+      if (columnConstraint?.raw_expr) {
+        addExpressionDependencies(columnConstraint.raw_expr, requires);
+      }
+    }
+
+    if (command?.subtype === "AT_ColumnDefault" && command.def) {
+      addExpressionDependencies(command.def, requires);
     }
 
     if (command?.subtype === "AT_ChangeOwner") {
--- a/packages/pg-topo/src/extract/extract-dependencies.ts
+++ b/packages/pg-topo/src/extract/extract-dependencies.ts
@@ -154,6 +154,25 @@ const extractCreateTableAsDependencies = (
   return { provides, requires };
 };
 
+const addConstraintExpressionDependencies = (
+  constraint: Record<string, unknown> | undefined,
+  requires: ObjectRef[],
+): void => {
+  if (!constraint) {
+    return;
+  }
+  for (const expressionNode of [
+    constraint.raw_expr,
+    constraint.cooked_expr,
+    constraint.exclusions,
+    constraint.where_clause,
+  ]) {
+    if (expressionNode) {
+      addExpressionDependencies(expressionNode, requires);
+    }
+  }
+};
+
 const extractAlterTableDependencies = (
   statementNode: Record<string, unknown>,
 ): ExtractDependenciesResult => {
@@ -189,9 +208,7 @@ const extractAlterTableDependencies = (
         provides.push(providedKey);
       }
     }
-    if (constraint?.raw_expr) {
-      addExpressionDependencies(constraint.raw_expr, requires);
-    }
+    addConstraintExpressionDependencies(constraint, requires);
 
     if (columnDefinition?.raw_default) {
       addExpressionDependencies(columnDefinition.raw_default, requires);
@@ -201,9 +218,7 @@ const extractAlterTableDependencies = (
       : [];
     for (const constraintItem of columnConstraints) {
       const columnConstraint = asRecord(asRecord(constraintItem)?.Constraint);
-      if (columnConstraint?.raw_expr) {
-        addExpressionDependencies(columnConstraint.raw_expr, requires);
-      }
+      addConstraintExpressionDependencies(columnConstraint, requires);
     }
 
     if (command?.subtype === "AT_ColumnDefault" && command.def) {
--- a/packages/pg-topo/test/alter-table-expression-dependencies.test.ts
+++ b/packages/pg-topo/test/alter-table-expression-dependencies.test.ts
@@ -1,0 +1,117 @@
+import { describe, expect, test } from "bun:test";
+import { analyzeAndSort } from "../src/analyze-and-sort";
+import { validateAnalyzeResultWithPostgres } from "./support/postgres-validation";
+
+const seededRandom = (seed: number): (() => number) => {
+  let state = seed >>> 0;
+  return () => {
+    state = (state * 1664525 + 1013904223) >>> 0;
+    return state / 4294967296;
+  };
+};
+
+const shuffleDeterministic = <T>(items: T[], seed: number): T[] => {
+  const random = seededRandom(seed);
+  const cloned = [...items];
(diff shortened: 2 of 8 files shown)


packages/pg-topo/src/extract/extract-dependencies.ts
```diff
@@ -188,6 +189,26 @@ const extractAlterTableDependencies = (
         provides.push(providedKey);
       }
     }
+    if (constraint?.raw_expr) {
+      addExpressionDependencies(constraint.raw_expr, requires);
```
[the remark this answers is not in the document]
xmh1011: Fixed in 530878f by walking the additional expression-bearing constraint fields (`cooked_expr`, `exclusions`, and `where_clause`) for ALTER TABLE constraints. Added runtime-backed coverage for EXCLUDE element expressions and EXCLUDE predicates.

Validation:
`bun run --cwd packages/pg-topo scripts/run-tests.ts test/alter-table-expression-dependencies.test.ts` -> 7 pass / 0 fail
`bun run format-and-lint` -> pass
`bun run check-types` -> pass

packages/pg-topo/src/extract/extract-dependencies.ts
```diff
@@ -154,6 +154,25 @@ const extractCreateTableAsDependencies = (
...
+const addConstraintExpressionDependencies = (
+  constraint: Record<string, unknown> | undefined,
+  requires: ObjectRef[],
+): void => {
+  if (!constraint) {
+    return;
+  }
+  for (const expressionNode of [
+    constraint.raw_expr,
+    constraint.cooked_expr,
```
avallete: **Blocking:** `cooked_expr` is dead code here. In raw-parser output it's only populated server-side during constraint transformation (and even then it's a serialized node *string*, not a JSON AST), so it never appears in SQL parsed by pg-topo and walking it can never produce a dependency. Please drop it from this list so the function only enumerates expression surfaces that can actually occur in parsed DDL.

---

packages/pg-topo/test/alter-table-expression-dependencies.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  const random = seededRandom(seed);
+  const cloned = [...items];
+  for (let index = cloned.length - 1; index > 0; index -= 1) {
+    const randomIndex = Math.floor(random() * (index + 1));
+    const current = cloned[index];
+    cloned[index] = cloned[randomIndex] as T;
+    cloned[randomIndex] = current as T;
+  }
+  return cloned;
+};
```
avallete: **Blocking:** this is now the fourth copy of `seededRandom`/`shuffleDeterministic` in the test tree (the others are in `test/function-body-dependency-chain.test.ts`, `test/support/randomized-runtime-analysis.ts`, and `test/support/fixture-regression.ts`). Rather than adding another duplicate, please extract a shared helper (e.g. `test/support/shuffle.ts`) and import it here. Migrating the three existing copies in the same PR would be welcome but isn't required — the ask is to stop the duplication from growing.

---

packages/pg-topo/src/extract/extract-dependencies.ts
```diff
@@ -188,6 +208,22 @@ const extractAlterTableDependencies = (
...
+
+    if (columnDefinition?.raw_default) {
+      addExpressionDependencies(columnDefinition.raw_default, requires);
+    }
+    const columnConstraints = Array.isArray(columnDefinition?.constraints)
+      ? columnDefinition.constraints
+      : [];
+    for (const constraintItem of columnConstraints) {
+      const columnConstraint = asRecord(asRecord(constraintItem)?.Constraint);
+      addConstraintExpressionDependencies(columnConstraint, requires);
```
avallete, line 244: **Minor (non-blocking):** `ADD COLUMN` still doesn't extract the column's `typeName` dependency or column-level `REFERENCES` edges from `ColumnDef.constraints` (the new loop only pulls *expressions*; the `CREATE_TABLE` path also calls `typeFromTypeNameNode` and `addForeignConstraintDependencies` for nested column constraints). I probed both cases and they currently pass anyway because `CREATE_TYPE`/`CREATE_TABLE` deterministically sort before `ALTER_TABLE` — so the missing edges are masked by class-tier ordering rather than truly tracked. Fine to leave for a follow-up, but worth either two extra lines here or a tracking note so the graph stays honest if class ordering ever changes.

---

avallete (changes requested):
Thanks for the thorough fix — I verified it locally: the regression tests fail at the test-only commit (0 pass / 5 fail) and the full pg-topo suite passes on the head (151 pass / 0 fail), with clean types and lint. The approach (routing `AlterTableCmd.def` expression surfaces through the existing `addExpressionDependencies` walker, mirroring the `CREATE_TABLE` path) is right, and the coverage beyond the two issue repros (`SET DEFAULT`, generated columns, `TYPE ... USING`, `EXCLUDE`) is appreciated.

Requesting changes for two small items flagged inline:

1. Drop the dead `cooked_expr` entry in `addConstraintExpressionDependencies`.
2. Extract the duplicated `seededRandom`/`shuffleDeterministic` test helpers into a shared `test/support` module instead of adding a fourth copy.

There's also one non-blocking inline note about `ADD COLUMN` type / column-level FK edges that can be a follow-up.

---

packages/pg-topo/src/extract/extract-dependencies.ts
xmh1011: Fixed in 2f0a247: dropped `cooked_expr` from the constraint expression surfaces. The ALTER TABLE walker now only enumerates raw-parser DDL expression fields that can actually be present here: `raw_expr`, `exclusions`, and `where_clause`.

packages/pg-topo/test/alter-table-expression-dependencies.test.ts
xmh1011: Fixed in fc663c0: added `test/support/shuffle.ts` and moved the new ALTER TABLE test plus the three existing deterministic-shuffle callers to that helper, leaving a single implementation.

packages/pg-topo/src/extract/extract-dependencies.ts
xmh1011: Addressed in 2f0a247 as part of this PR: `ALTER TABLE ... ADD COLUMN` now extracts the column `typeName`, column-level `CONSTR_FOREIGN` dependencies, and column-level primary/unique key providers. Added focused regressions that assert the ALTER TABLE requires set includes the referenced type, referenced table, and referenced key.

avallete (approved):
Thank you for your contribution !

PR #290, opened 2026-06-11
fix(pg-topo): support create rule dependencies

### LINKED ISSUE #284: pg-topo: CREATE RULE is treated as unsupported and misses expression dependencies
## Summary

`pg-topo` currently treats `CREATE RULE` statements as `UNKNOWN`. When the rule `WHERE` expression references a function, the static sorter does not extract that dependency and can order the rule before the function it calls.

This is a separate issue class from the currently reported pg-topo issues:

* supabase/pg-toolbelt#281 covers expression dependencies inside `ALTER TABLE` subcommands.
* supabase/pg-toolbelt#282 covers `CREATE TYPE ... AS RANGE` / `CreateRangeStmt` coverage.
* supabase/pg-toolbelt#283 covers `ALTER PUBLICATION` / `ALTER SUBSCRIPTION` coverage.
* This issue is about the parser's `RuleStmt` AST root and rule expression extraction.

Current `main`: `9f01826b00b64e7f9628844ecb6f69a07f53c1f8`

## Repro

Input statements, intentionally shuffled:

```sql
create rule block_invalid AS
  ON INSERT TO app.items
  WHERE app.is_invalid(NEW.flag)
  DO INSTEAD NOTHING;

create function app.is_invalid(value boolean)
returns boolean
language sql
immutable
as $$ select value is false $$;

create table app.items(id int primary key, flag boolean);
create schema app;
```

## Actual result

`pg-topo` classifies the rule as `UNKNOWN` and orders it before the function used by the rule expression:

```json
{
  "ordered": [
    {
      "class": "CREATE_SCHEMA",
      "sql": "create schema app;"
    },
    {
      "class": "CREATE_TABLE",
      "sql": "create table app.items(id int primary key, flag boolean);"
    },
    {
      "class": "UNKNOWN",
      "sql": "create rule block_invalid AS ON INSERT TO app.items WHERE app.is_invalid(NEW.flag) DO INSTEAD NOTHING;"
    },
    {
      "class": "CREATE_FUNCTION",
      "sql": "create function app.is_invalid(value boolean) returns boolean language sql immutable as $$ select value is false $$;"
    }
  ],
  "diagnostics": [
    {
      "code": "UNKNOWN_STATEMENT_CLASS",
      "message": "Unsupported statement AST root 'RuleStmt'."
    }
  ]
}
```

Runtime validation then fails when PostgreSQL executes the rule before the function:

```text
code: RUNTIME_EXECUTION_ERROR
message: function app.is_invalid(boolean) does not exist
sqlstate: 42883
missingObjectKey: function:app:is_invalid:(boolean)
producerStatementId: <input:3>:0
```

## Expected result

The rule should be recognized as a supported post-table object and ordered after both:

1. the target relation `app.items`; and
2. objects referenced by its expressions, such as `app.is_invalid(boolean)` in the rule `WHERE` clause.

Expected order for the repro:

```text
CREATE_SCHEMA
CREATE_TABLE
CREATE_FUNCTION
CREATE_RULE
```

No `UNKNOWN_STATEMENT_CLASS` diagnostic should be emitted for `RuleStmt` in this case.

## Parser shape

The parser exposes structured refs that should be enough for static extraction:

```json
{
  "RuleStmt": {
    "relation": {
      "schemaname": "app",
      "relname": "items"
    },
    "rulename": "block_invalid",
    "whereClause": {
      "FuncCall": {
        "funcname": ["app", "is_invalid"],
        "args": ["NEW.flag"]
      }
    },
    "event": "CMD_INSERT",
    "instead": true
  }
}
```

## Root cause

`packages/pg-topo/src/classify/classify-statement.ts` does not map `RuleStmt`, so the statement gets the `UNKNOWN` class. Because it never reaches a rule-specific extractor, `extractDependenciesForStatement` does not emit dependencies on:

* the target table/view relation; and
* objects referenced by the rule `whereClause` and rule action(s), when present.

The default `UNKNOWN` phase also lets the rule sort before routine creation when there is no extracted edge.

## Suggested fix direction

1. Add a `CREATE_RULE` statement class and map `RuleStmt` to it.
2. Put `CREATE_RULE` in the `post_data` phase, near triggers/policies, because rules depend on an existing relation.
3. Add dependency extraction for `RuleStmt`:
   * provide a stable rule ref, for example `rule:app:items.block_invalid` if `rule` becomes an object kind, or skip `provides` initially if the model does not need rule comments/ownership yet;
   * require the target relation from `RuleStmt.relation`;
   * call the existing `addExpressionDependencies(...)` helper on `RuleStmt.whereClause`;
   * inspect `RuleStmt.actions` when present and add expression dependencies from those action query nodes as well.
4. Add focused pg-topo tests for:
   * `CREATE RULE` with a function in `WHERE` ordered after the function and table;
   * a simple rule without expressions ordered after its relation;
   * no `UNKNOWN_STATEMENT_CLASS` diagnostic for `RuleStmt`.

## Local validation

The focused repro was run twice with the same result: `RuleStmt` is unknown and runtime validation fails with `42883` because the function is ordered after the rule.

```bash
./node_modules/.bin/bun run --cwd packages/pg-topo scripts/run-tests.ts \
  --test-name-pattern "CREATE RULE" \
  test/deep-probe.test.ts
```

I can work on a fix for this if the approach above sounds acceptable.

cc @avallete

xmh1011:
## Summary

Closes #284.

This adds pg-topo support for `CREATE RULE` statements by:

- classifying `RuleStmt` as `CREATE_RULE`
- adding `rule` as an object kind
- extracting the created rule object and target relation
- extracting dependencies from the rule `WHERE` predicate
- extracting dependencies from rule action queries with the existing expression traversal

This keeps the implementation focused on static ordering and avoids adding new graph complexity beyond the rule's actual object/expression dependencies.

## RED evidence

Before the fix, the focused rule regressions failed with `UNKNOWN_STATEMENT_CLASS` for `RuleStmt`, so pg-topo could not order rule predicate or action-query dependencies.

## Validation

```bash
bun run test:pg-topo
```

Result: `146 pass / 0 fail`, including runtime-backed rule `WHERE` and action-query dependency regressions.

```bash
bun run format-and-lint
bun run check-types
bun run knip
```

Result: passed. `knip` only reported existing configuration hints.


avallete (approved):
LGTM ! Thank's for your contribution !

packages/pg-topo/src/extract/extract-dependencies.ts
```diff
@@ -375,6 +375,41 @@ const extractTriggerDependencies = (
...
+      typeof statementNode.rulename === "string"
+        ? statementNode.rulename
+        : "rule";
+    provides.push(
+      createObjectRefFromAst(
+        "rule",
+        `${relationRef.name}.${ruleName}`,
+        relationRef.schema ?? DEFAULT_SCHEMA,
+      ),
+    );
```
avallete, line 397: Now that rules are first-class providers, `COMMENT ON RULE r ON app.users` still won't link to this provide: `objectKindFromObjType` in `shared-refs.ts` has no `OBJECT_RULE` case, and the `relation.name` identity special-case in `objectFromNameParts` covers only `trigger`/`policy`. So a rule comment extracts no requires and can sort before the rule exists.

This was equally broken before this PR (the rule itself was `UNKNOWN`), so it's not a regression — but it's the natural follow-up. Fine to either add the `OBJECT_RULE` mapping + extend the `objectFromNameParts` special case here, or file a tracking issue.

---

packages/pg-topo/src/extract/extract-dependencies.ts
```diff
@@ -375,6 +375,41 @@ const extractTriggerDependencies = (
...
+    requires.push(relationRef);
+  }
+
+  addExpressionDependencies(statementNode.whereClause, requires);
+
+  const actions = Array.isArray(statementNode.actions)
+    ? statementNode.actions
+    : [];
+  for (const action of actions) {
+    addExpressionDependencies(action, requires);
```
avallete, line 407: Informational, no change needed: an action that references `OLD`/`NEW` as a range table (e.g. `DO INSTEAD SELECT * FROM old`) parses as a plain `RangeVar`, so this would emit a false `table:public:old` require and an `UNRESOLVED_DEPENDENCY` diagnostic. Very rare in practice and consistent with how the rest of the extractor treats `RangeVar`s — only worth special-casing if it ever shows up.

---

packages/pg-topo/test/statement-coverage.test.ts
```diff
@@ -159,6 +160,90 @@ describe("statement coverage", () => {
     expect(viewIndex).toBeGreaterThan(fnIndex);
   });
 
+  test("orders create rule after target table and WHERE predicate function", async () => {
```
avallete, line 163: #284's test plan also listed "a simple rule without expressions ordered after its relation". Both tests here involve functions, so the bare relation-only edge is only covered transitively. A cheap one-liner test (e.g. `create rule r as on delete to app.users do instead nothing;` + the table + schema) would complete the issue's acceptance criteria.

---

avallete (commented):
Solid PR — the implementation closely follows the design suggested in #284 and mirrors the existing trigger/policy extractor patterns. Verified the integration points against the codebase; no correctness issues found:

- The `schema:relation.rulename` provide identity matches the trigger/policy pattern, and rule names are unique per relation, so the identity is correct.
- Passing whole action nodes to `addExpressionDependencies` works: the visitor walks the full subtree and picks up `RangeVar` (action-target tables), `FuncCall`, and `TypeName` nodes.
- Rules on views are handled despite the hard-coded `"table"` require kind, since `isKindCompatible` resolves `table` requires against `view`/`materialized_view` provides.
- The weight renumbering in `topo-sort.ts` is a pure insertion — all relative tie-break orders preserved.
- Changeset, conventional title, RED evidence, and runtime-validated tests with shuffled inputs all check out.

Left three minor, non-blocking comments inline: a `COMMENT ON RULE` linking gap that predates this PR (follow-up worthy), the missing relation-only test case from #284's test plan, and one informational edge case. LGTM once CI is green.

---

packages/pg-topo/src/extract/extract-dependencies.ts
xmh1011: Verified against merged PR #290 (`436b3d1` on `main`): `CREATE RULE` support is merged, but `OBJECT_RULE` and rule relation-qualified identity are not. Since this PR is already merged, I am not pushing to the merged branch; this should be handled as a post-merge follow-up for `COMMENT ON RULE ... ON relation` ordering.

packages/pg-topo/src/extract/extract-dependencies.ts
xmh1011: No code change for this thread. I verified this is still the extractor-wide `RangeVar` behavior, so `OLD`/`NEW` used as action range tables can produce an unresolved table reference. It is not specific to the new `CREATE_RULE` path, and I am not changing the already-merged PR branch.

packages/pg-topo/test/statement-coverage.test.ts
xmh1011: Verified against merged `main` (`436b3d1`): the PR includes rule tests with predicate/action dependencies, but not the simple relation-only rule case from #284. Since the PR is already merged, I am not pushing to this branch; the cheap `create rule r as on delete to app.users do instead nothing` coverage should be added as a post-merge follow-up.

avallete: Handling everything here: supabase/pg-toolbelt#294

xmh1011: Confirmed. #294 covers the post-merge follow-up for `COMMENT ON RULE` and the direct relation-only rule test, so I will not push anything further to this already-merged #290 branch.

Issue #244, opened 2026-05-02
refactor(pg-delta): move execution phase onto BaseChange as `change.phase`


avallete:
## Summary

Replace the standalone `getExecutionPhase(change)` helper in `packages/pg-delta/src/core/sort/utils.ts` with a `phase` getter on `BaseChange`. Rename the enum value `"create_alter_object"` → `"forward"`. Keep behavior identical.

This concentrates \"which execution pass does this change belong to?\" on the change itself (high locality), removes a layering violation where generic sort code reaches into concrete `AlterTableAlterColumn*` classes via `instanceof`, and makes the classification total (exhaustive over `ChangeOperation`, no silent fallback).

## Motivation

Today the same question — \"is this a destructive/reverse-order change, or a forward one?\" — is answered in at least four places with three subtly-different rules:

1. `packages/pg-delta/src/core/sort/utils.ts` → `getExecutionPhase(change)` with a mix of `operation` / `scope` / `drops` / `instanceof` checks and a `Safe default` fallback at the bottom.
2. `packages/pg-delta/src/core/catalog.diff.ts:220-235` derives \"is this a top-level object drop?\" via `operation === \"drop\" && scope === \"object\"`.
3. `packages/pg-delta/src/core/sort/logical-sort.ts` and `sort/custom-constraints.ts` pattern-match on `(operation, scope)` pairs.
4. `packages/pg-delta/src/core/plan/create.ts:365` has `hasRoutineChanges` as yet another \"predicate over a change\" — left out of this issue, but flagged as the next candidate (see _Out of scope_).

The `instanceof` leaks are the worst smell: `sort/utils.ts` imports concrete `AlterTableAlterColumn*` classes from `objects/table/changes/` to decide their phase. That's a generic sort module reaching across the codebase into a specific object family. With a getter on the change class, that knowledge colocates with the class itself.

## Design (settled)

All load-bearing decisions have been made — no further grilling needed.

| Decision | Choice |
|---|---|
| Scope | **A (narrow)** — phase only. `operation`/`scope` stay as-is; they're part of the user-facing DSL wire format (see `integrations/supabase.ts`, `integrations/filter/dsl.ts`, `Plan.filter` serialization). |
| Interface shape | **Getter on `BaseChange`** with a default implementation + selective overrides. Not strictly abstract — ~150 concrete change classes would need identical 3-line getters otherwise. |
[...]

PR #293, opened 2026-06-11
feat(pg-delta)!: replace flat plan statements with execution-aware migration units

### LINKED ISSUE #262: pg-delta: ALTER TYPE ADD VALUE and dependent DDL must run in separate transaction scopes (55P04)
## Bug report

pg-delta generates `ALTER TYPE ... ADD VALUE` and DDL that references the new enum value in the same migration transaction.

PostgreSQL rejects this with error `55P04` because newly added enum values are not visible until the transaction is committed.

- [x] I confirm this is a bug with Supabase, not with my own application.
- [x] I confirm I have searched the Docs, GitHub Discussions, and Discord.

---

## Describe the bug

When a migration both:

- adds a new enum value
- and references that value in subsequent DDL

pg-delta emits all statements in the same transaction.

For example:

```sql
ALTER TYPE public.user_role ADD VALUE 'store';

ALTER TABLE public.profiles
ALTER COLUMN role SET DEFAULT 'store'::public.user_role;
```

PostgreSQL rejects this because enum values added via `ALTER TYPE ... ADD VALUE` are only usable after transaction commit.

This issue affects statements such as:

- `ALTER COLUMN ... SET DEFAULT`
- `ADD CONSTRAINT ... CHECK (...)`

This does **not** apply to `CREATE TYPE AS ENUM (...)`, where all enum values exist at type creation time.

---

## PostgreSQL behavior

From the PostgreSQL documentation:

> New enum values must be committed before they can be used.

---

### Case 1: ADD VALUE + SET DEFAULT (55P04)

Schema in both databases:

```sql
CREATE TYPE public.user_role AS ENUM ('admin', 'user');

CREATE TABLE public.profiles (
  id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
  name text NOT NULL,
  role public.user_role DEFAULT 'user'
);
```

Change to target database:

```sql
ALTER TYPE public.user_role ADD VALUE 'store';

ALTER TABLE public.profiles
ALTER COLUMN role SET DEFAULT 'store';
```

Generated SQL by pg-delta:

```sql
ALTER TYPE public.user_role ADD VALUE 'store' AFTER 'user';

ALTER TABLE public.profiles
ALTER COLUMN role SET DEFAULT 'store'::public.user_role;
```

Apply result:

```text
ERROR: 55P04: unsafe use of new value "store" of enum type user_role
HINT: New enum values must be committed before they can be used.
```

---

### Case 2: ADD VALUE + ADD CHECK constraint (55P04)

Schema in both databases:

```sql
CREATE TYPE public.order_status AS ENUM ('pending', 'shipped');

CREATE TABLE public.orders (
  id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
  status public.order_status DEFAULT 'pending'
);
```

Change to target database:

```sql
ALTER TYPE public.order_status ADD VALUE 'delivered';

ALTER TABLE public.orders ADD CONSTRAINT status_check
CHECK (status IN ('pending', 'shipped', 'delivered'));
```

Generated SQL by pg-delta:

```sql
ALTER TYPE public.order_status ADD VALUE 'delivered' AFTER 'shipped';

ALTER TABLE public.orders ADD CONSTRAINT status_check
CHECK (
  status = ANY (
    ARRAY[
      'pending'::public.order_status,
      'shipped'::public.order_status,
      'delivered'::public.order_status
    ]
  )
);
```

Apply result:

```text
ERROR: 55P04: unsafe use of new value "delivered" of enum type order_status
HINT: New enum values must be committed before they can be used.
```

---

## Expected behavior

pg-delta should ideally ensure that `ALTER TYPE ... ADD VALUE` statements are executed in a separate transaction scope before any DDL that references the new enum value.

One possible approach could be emitting a transaction boundary between the enum alteration and the dependent DDL. For example:

```sql
ALTER TYPE public.user_role ADD VALUE 'store' AFTER 'user';

COMMIT;

ALTER TABLE public.profiles
ALTER COLUMN role SET DEFAULT 'store'::public.user_role;
```

---

## System information

- OS: Ubuntu 24.04.3 LTS
- pg-delta version: 1.0.0-alpha.25
- PostgreSQL version: 17.6
- Test framework: Roundtrip Fidelity Tests
- Browser/Client: N/A (CLI tool)

avallete:
## Summary

Replace the flat `plan.statements` array with an ordered list of `MigrationUnit`s that carry explicit transaction boundaries and execution semantics. This fixes PostgreSQL error 55P04 ("unsafe use of new value of enum type") when migrations add enum values and immediately use them, and correctly handles statements that cannot run inside transaction blocks.

Closes #262
Supersedes #277 (ground-up redesign of the same feature; see that PR's review threads for the motivation: trait-based classification instead of regex over rendered SQL, no per-statement effect machinery, single source of truth instead of triple storage).

## Key Changes

- **New execution planning** (`execution.ts`): `buildExecutionPlan()` groups sorted changes into transaction-aware migration units based on `nonTransactional` and `commitBoundary` traits declared on change classes, never inferred from SQL.

- **Plan structure v2**: Plans now carry `units: MigrationUnit[]` (each with `transactionMode`, `reason`, and `statements`) plus `sessionStatements: string[]`, replacing the flat `statements` array. Legacy v1 plans are normalized on-the-fly by `normalizePlan()`.

- **Commit-visibility boundaries**: New `CommitBoundaryReason` type (`"enum_value_visibility"`) marks changes whose effects only become usable after COMMIT. `AlterEnumAddValue` declares this boundary; subsequent statements run in a separate transactional unit. The rule is conservative by construction: *any* statement after the producer run is pushed past the commit — no consumer detection.

- **Non-transactional statements**: Changes can declare `nonTransactional = true` for statements PostgreSQL rejects inside `BEGIN`/`COMMIT`. These emit single-statement units with `transactionMode: "none"`.

- **Subscription traits** (`subscription.*.ts`): `CreateSubscription` (existing replication slot keeps the `connect = true` default), `AlterSubscriptionSetPublication` (implicit `refresh = true` when enabled), and `DropSubscription` (associated slot) now declare `nonTransactional`.

- **Plan rendering** (`render.ts`):
  - `renderPlanSql()`: single script with unit headers and explicit BEGIN/COMMIT per transactional unit
  - `renderPlanFiles()`: numbered SQL files (one per unit) for independent execution
  - `flattenPlanStatements()`: flatten back to the ordered statement list, session statements included (execution context lost)

- **Plan application** (`apply.ts`): `applyPlan()` executes units in order on a single session, wrapping transactional units in explicit BEGIN/COMMIT and running non-transactional units without a wrapper. Failures do not roll back already-committed units; the result includes `failedUnitIndex` and `completedUnits` for diagnostics.

- **Typed cycle errors**: `sortChanges` now throws `UnorderableCycleError` (exported) carrying the offending changes in `error.cycle`, instead of a plain `Error` callers had to string-match; the CLI prints the diagnostic with a report hint.

- **CLI**: new `pgdelta plan --output-dir <dir>` flag writes the numbered migration files via `renderPlanFiles()`.

## Implementation Details

- Execution semantics are **declared on change classes** via the `nonTransactional` and `commitBoundary` getters (siblings of `creates`/`drops`/`invalidates`/`requires`, consistent with the settled design of #244), not inferred from rendered SQL. This keeps the planner independent of serialization details and integrations.
- Session statements (SET ROLE, SET check_function_bodies) are separated from migration units and repeated in each rendered file, since each file may execute in its own session.
- The planner groups consecutive producers of the same `commitBoundary` kind into one unit; a different boundary kind or a non-producer flushes the current unit and starts a new one whose `reason` is the producer's token.
- The dead `BaseChange.changeId` getter (zero consumers; embedded the full serialized SQL) is deleted rather than persisted into plan JSON.
- Plan JSON is now `version: 2`. Legacy v1 files still deserialize (leading SET statements become session statements, the rest one transactional unit — faithful to v1's single implicit transaction); v2 files are not readable by older pg-delta. See the changeset for the full migration guide.

https://claude.ai/code/session_011E5qkRMxoTcVgTYfAX5dFk


packages/pg-delta/src/core/objects/subscription/changes/subscription.create.ts
```diff
@@ -23,6 +23,12 @@ export class CreateSubscription extends CreateSubscriptionChange {
     return [stableId.role(this.subscription.owner)];
   }
 
+  // With an existing replication slot, serialize() keeps the connect = true
+  // default, which PostgreSQL rejects inside a transaction block.
+  override get nonTransactional() {
+    return this.subscription.replication_slot_created;
```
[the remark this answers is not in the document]
avallete: Confirmed and fixed in 001cf41 — `serialize()` now emits `create_slot = false` when `replication_slot_created` is true, keeping the `connect = true` default so the existing slot is looked up and reused.

One correction to the framing, verified empirically against PostgreSQL 17: the 25001 transaction-block gate is on `create_slot = true`, not on `connect` — `connect = true, create_slot = false` runs fine inside a `BEGIN`/`COMMIT` block (and works same-cluster). So with the serializer fixed, no emitted `CREATE SUBSCRIPTION` form is non-transactional anymore, and the `nonTransactional` override on `CreateSubscription` is removed rather than kept: these plans now stay fully transactional.

Note the apply leg still requires manual intervention by design: extraction redacts `conninfo` to the `__CONN_*__` placeholder (credentials), so the emitted statement can't reach a publisher until real connection values are substituted. Covered by a unit regression on the serializer and an integration test asserting the transactional plan shape against a real `pg_replication_slots` row.

---

packages/pg-delta/src/core/plan/render.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  const lines: string[] = [
+    `-- Migration unit ${index + 1}: ${unitName(unit.reason)}`,
+    `-- Transaction mode: ${unit.transactionMode}`,
+    `-- Boundary reason: ${unit.reason}`,
+    "",
+  ];
+
+  if (plan.sessionStatements.length > 0) {
+    lines.push(renderStatements(plan.sessionStatements));
```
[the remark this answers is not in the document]
avallete: Correct observation, but the suggested separation can't fix it: PostgreSQL executes **every** statement of a multi-command simple-query string inside an implicit transaction block, so a `transactionMode: "none"` statement can never run from a single query string — with or without the leading `SET`s, and even with interleaved `COMMIT`s (the remaining statements just open a new implicit block). The `SET` placement isn't the failing ingredient; single-string execution is.

The two supported execution paths are unaffected: `applyPlan` executes statement-by-statement on one session, and `psql -f` splits statements (which is also why the session `SET`s must stay in the file — they apply to the psql session before the payload).

Fixed in c4c27db by making the contract explicit instead of reshuffling statements: non-transactional units now carry a "run statement-by-statement (do not use psql -1 or a single multi-statement query string)" header in rendered output, and `docs/api.md` + the changeset document that rendered scripts target statement-splitting runners, with `applyPlan` as the programmatic path.

---

packages/pg-delta/src/core/plan/render.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+function trimTerminator(statement: string): string {
+  const trimmed = statement.trim();
+  let end = trimmed.length;
+  while (end > 0 && trimmed.charCodeAt(end - 1) === 59) {
+    end--;
+  }
+  return trimmed.slice(0, end);
+}
```
avallete, line 153: **question**

Why do we need to trim before re-adding the delimiter (plus the extra ;) ?

packages/pg-delta/tests/integration/subscription-operations.test.ts
```diff
@@ -230,5 +232,115 @@ for (const pgVersion of POSTGRES_VERSIONS) {
...
+        expect(unit.statements).toHaveLength(1);
+        const [createStatement] = unit.statements;
+        expect(createStatement).toStartWith(
+          "CREATE SUBSCRIPTION sub_with_slot",
+        );
+        expect(createStatement).toContain("create_slot = false");
+        // connect must stay at its default (true) so the existing slot is
+        // looked up on the publisher rather than skipped.
+        expect(createStatement).not.toContain("connect = false");
+        expect(createStatement).toContain("slot_name = 'sub_existing_slot'");
```
avallete, line 288: **nipcik**

I think we could test the "apply leg" by just hardcoding a replace over the produced `createStatement` then we can try to apply it and see if it passes.

packages/pg-delta/tests/integration/type-operations.test.ts
```diff
@@ -26,6 +28,77 @@ for (const pgVersion of POSTGRES_VERSIONS) {
...
+        expect(result.plan.units[1].reason).toBe("enum_value_visibility");
+
+        const applied = await applyPlan(result.plan, db.main, db.branch);
+        expect(applied.status).toBe("applied");
+        if (applied.status !== "applied") throw new Error("expected applied");
+        expect(applied.warnings).toBeUndefined();
+        const after = await createPlan(db.main, db.branch);
+        expect(after).toBeNull();
+      }),
+    );
```
avallete, line 93: **nipick**

Can't that be tested using the recommended "roundtrip" test ?

packages/pg-delta/src/core/plan/render.ts
avallete: Because unit `statements` aren't guaranteed terminator-free: pg-delta's own serializers never emit a trailing `;`, but plan JSON is a persisted artifact — legacy v1 files, hand-built units (our own tests do this), or user-edited plans may already carry one, and joining those blindly would render `stmt;;`. Trimming first (the loop also collapses repeated `;`) makes rendering canonical regardless of where the plan came from. Added a doc comment on `trimTerminator` in a59651d so this doesn't need re-asking.

---

packages/pg-delta/tests/integration/subscription-operations.test.ts
avallete: Good idea — done in a59651d. The test now substitutes real connection values over the redacted conninfo (`createStatement.replace(/CONNECTION '[^']*'/, ...)`, exactly what a user does) and executes the result inside an explicit `BEGIN`/`COMMIT` against the same cluster, then asserts the subscription exists on the target. That pins both claims in CI: the `create_slot = false` form really is transactional, and it reuses the existing slot instead of recreating it.

---

packages/pg-delta/tests/integration/type-operations.test.ts
avallete: Yes — converted in a59651d (both enum tests and the dollar-quote one in function-operations). `roundtripFidelityTest` applies via `applyPlan` internally, so the per-unit transaction semantics are still exercised; I added an `assertPlan` hook to the helper for the migration-unit shape assertions the roundtrip couldn't express. One wrinkle: the branch *setup* has to stay as two separate queries rather than `testSql`, because the helper runs `testSql` as a single multi-statement query — which itself hits 55P04 while preparing the branch (noted in a test comment).

---

avallete (commented):
cc @jgoux LGTM

jgoux (approved):

PR #302, opened 2026-06-20
feat(postgrest-typegen): add package with multi-language code generation

avallete:
## Summary

Introduces `@supabase/postgrest-typegen`, a new package that provides type generation for PostgREST from PostgreSQL schemas. This is the type-generation engine extracted from postgres-meta, offering a clean separation between introspection and code generation with support for TypeScript, Go, Python, and Swift.

## Key Changes

- **New package structure**: `packages/postgrest-typegen/` with complete introspection and generation pipeline
- **Database introspection**: SQL-based introspection that mirrors postgres-meta's approach, extracting schemas, tables, views, columns, functions, types, and relationships
- **Multi-language code generation**:
  - TypeScript: Full `Database` type with Row/Insert/Update operations, relationships, and helper types
  - Go: Struct generation with JSON tags for Select/Insert/Update operations
  - Python: Pydantic models and TypedDict definitions with proper type hints
  - Swift: Codable structs with CodingKeys and protocol conformances
- **Type validation**: ArkType-backed schema validation for `GeneratorMetadata` contract, ensuring compile-time equivalence with postgres-meta's types
- **Comprehensive test coverage**:
  - Unit tests for each generator with inline snapshots
  - SQL introspection tests pinning exact query output
  - Integration tests against real PostgreSQL containers
  - Parity tests validating output against expected fixtures
- **Fixture database**: Two SQL files (`00-init.sql`, `01-memes.sql`) providing test data including tables, views, enums, functions, and relationships

## Notable Implementation Details

- **Verbatim port from postgres-meta**: SQL builders and generation templates are ported directly from postgres-meta to maintain byte-parity and ensure compatibility
- **Pluggable metadata contract**: `GeneratorMetadata` is the central interface allowing different introspection sources (SQL, API, etc.) to feed the same generators
- **Relationship expansion**: Handles both table-to-table and view-to-table relationships, with support for detecting one-to-one relationships
- **Normalized output**: All generators produce formatted, validated output (TypeScript via Prettier, others via built-in formatting)
- **Configuration options**: Each generator accepts options (e.g., `detectOneToOneRelationships`, `postgrestVersion`, `defaultSchema` for TypeScript; `accessControl` for Swift)

## Testing

- Run all tests: `bun run test:postgrest-typegen`
- Integration tests require Docker (PostgreSQL containers via testcontainers)
- Parity tests validate output against expected snapshots from real database introspection

https://claude.ai/code/session_01B53a2ayu9d16u821FQ5Qqz


avallete:
## Re: Codex automated review (13 comments)

Thanks for the thorough pass. Triage below. **TL;DR:** 12 of the 13 describe behavior **ported verbatim** from postgres-meta (the `supabase gen types` engine), so they are pre-existing — not regressions introduced here — and fixing them in this PR would break the port's explicit byte-parity constraint. The 1 comment that touches *new* code is fixed in `74e498f`.

### Why most are intentionally deferred

This package is a **byte-for-byte extraction** of postgres-meta's type-generation engine. The stated constraint (README / package `CLAUDE.md` / changeset) is: **generator output must stay byte-identical to postgres-meta's templates until parity is validated and released — no behavioral "improvements" during the port.** This is enforced by a parity test that diffs generated output against fixtures captured from postgres-meta, and it currently passes. I also diffed the one SQL file flagged (`functions.sql.ts`) against postgres-meta and confirmed the logic is identical (only quote-style/formatting differs).

So each item below reproduces identically against today's `supabase gen types`. "Fixing" them in the package now would (a) diverge from postgres-meta and (b) fail the parity gate. The correct home is **upstream in postgres-meta first, then re-port** (or as tracked post-parity follow-ups in both repos):

- `typescript.ts:151` — cross-schema relationships dropped
- `typescript.ts:673` / `python.ts:174` — generated columns still accepted in Insert
- `typescript.ts:871` — `DefaultSchema` → `never` when default schema absent
- `go.ts:37` / `python.ts:22` — foreign tables omitted from Go/Python output
- `go.ts:213` — enum attributes unresolved in Go composites
- `python.ts:89` — Python enums not filtered to selected schemas
- `python.ts:230` / `swift.ts:187` — enum labels not escaped
- `swift.ts:396` — cross-schema type referenced but not emitted

These are legitimate observations — they're just out of scope for a parity-preserving extraction.

### The one fixed here

- **`functions.sql.ts:140` (null `has_default` for OUT/TABLE args)** — fixed in `74e498f`. This is the only comment touching *new* code: `introspect()` legitimately emits `has_default: null` for the OUT columns of `RETURNS TABLE` / OUT-arg functions (the SQL sizes `arg_has_defaults` from the input-arg count), but the new opt-in `parseGeneratorMetadata` validator typed the field as a plain `boolean` and rejected it. Widened to `boolean | null`. Generator output is unaffected (the generators already treat `has_default` truthily, so `null` and `false` behave identically), so parity stays green. Covered by a unit regression plus an integration assertion against the real `function_returning_table` fixture.

I won't 👍/👎 each inline thread to avoid noise — happy to open tracking issues for the deferred items if that's preferred.

---

avallete:
## Re: second Codex pass (commit `74e498f`)

This pass surfaced 6 more suggestions — all in the **verbatim-ported** generation/SQL files (`types.sql.ts`, `go.ts`, `typescript.ts`, `python.ts`), none of which the fix commit touched (it only changed `src/types.ts` + tests). So they fall under the same rationale as [my earlier triage](https://github.com/supabase/pg-toolbelt/pull/302#issuecomment-4757271698): they reproduce identically against postgres-meta's `supabase gen types` today and are deferred to preserve the port's byte-parity gate (fix upstream first, then re-port).

For the record, this batch:

- `types.sql.ts:51` — foreign-table (`relkind='f'`) row types excluded from `types`
- `go.ts:171` — `omitempty` missing on optional Go write fields (P1)
- `typescript.ts:1094` — materialized-view row types not resolved
- `python.ts:116` — references un-emitted classes for non-enum/non-composite user types
- `python.ts:398` — `Json[Any]` rejects already-decoded JSON row values
- `python.ts:458` — Python-keyword column names not escaped

All legitimate, all pre-existing, none introduced by this PR. Happy to capture them (plus the first 12) as upstream/post-parity tracking issues if you'd like a single place to follow them.

---

avallete:
## Re: third Codex pass (commit `f9c14a8`)

Same disposition as the [earlier](https://github.com/supabase/pg-toolbelt/pull/302#issuecomment-4757271698) [triages](https://github.com/supabase/pg-toolbelt/pull/302#issuecomment-4757316740): all 11 suggestions land in the **verbatim-ported** generators (`typescript.ts`, `swift.ts`, `python.ts`, `go.ts`) — none touch new code — so they reproduce identically against postgres-meta's `supabase gen types` today and are deferred to keep the port's byte-parity gate (fix upstream first / post-parity). This batch: updatable-view write types (Go/Python/Swift), identifier & keyword escaping (Go/Python/Swift), Python alias keys, public Swift initializers, and cross-schema RPC/computed-field resolution by relation id.

These are good follow-ups; flagging for the consolidated upstream/post-parity list rather than fixing in this extraction PR.

---

spydon:
A few things that could be good to consider before merging this:

1. A version field on the serialized contract. Already raised in my comment on their RFC. Without it, out-of-process consumers like `supabase_typegen` can only detect shape changes, not semantic ones. This is the single change that most affects how safely our generator can evolve.

2. Schema-qualify enum types on columns. `PostgresColumn.format` carries only the bare enum type name, which is why `supabase_typegen` has a documented limitation: two schemas defining same-named enums are ambiguous, and a column using the other schema's enum silently resolves to the wrong value set (a runtime `ArgumentError` on valid data). They own the introspection SQL now, so adding a `type_schema` field (or qualifying format for user-defined types) fixes it at the root for every consumer. This is the only known correctness gap we cannot fix on our side.

3. Reconsider dropping primary_keys. Their contract removed it because the four existing templates don't use it, but it's catalog information the old document carried, and planned Dart features (`upsert` helpers with `onConflict` defaults, foreign-key relationship getters that want the target's key) will need it. Cheap to keep, expensive to retrofit once external consumers pin the shape.

spydon:
Superseded by https://github.com/supabase/sdk/pull/94

PR #313, opened 2026-06-26
fix(pg-delta): order leading enum value additions

### LINKED ISSUE #304: pg-delta: multiple leading enum values generate ADD VALUE before a non-existent label
## Bug report

`pg-delta` generates invalid `ALTER TYPE ... ADD VALUE` SQL when a target enum inserts a contiguous run of two or more new values before the first existing label.

This is different from supabase/pg-toolbelt#262: it is not about using a newly added enum value in dependent DDL before a commit. The generated enum DDL itself references an anchor label that does not exist yet.

- [X] I confirm this is a bug with Supabase, not with my own application.
- [X] I confirm I have searched the Docs, GitHub Discussions, and Discord.

## To Reproduce

Source database:

```sql
CREATE SCHEMA probe;
CREATE TYPE probe.status AS ENUM ('c');
```

Target database:

```sql
CREATE SCHEMA probe;
CREATE TYPE probe.status AS ENUM ('a', 'b', 'c');
```

On current `main` (`c06f081`, `@supabase/pg-delta@1.0.0-alpha.30`), `createPlan` produces one transactional unit:

```sql
ALTER TYPE probe.status ADD VALUE 'a' BEFORE 'b';
ALTER TYPE probe.status ADD VALUE 'b' AFTER 'a';
```

Applying that plan fails immediately:

```text
ERROR 22023: "b" is not an existing enum label
```

The failure is deterministic because the first statement tries to insert `a` before `b`, but `b` is also new and has not been added yet.

## Additional cases tested

These cases were validated against a temporary PostgreSQL 16.13 cluster using `createPlan` + `applyPlan` from `origin/main`:

| Source labels | Target labels | Generated shape | Result |
| -- | -- | -- | -- |
| `('b')` | `('a', 'b')` | `ADD 'a' BEFORE 'b'` | applies |
| `('c')` | `('a', 'b', 'c')` | `ADD 'a' BEFORE 'b'; ADD 'b' AFTER 'a'` | fails, `22023` |
| `('d')` | `('a', 'b', 'c', 'd')` | `ADD 'a' BEFORE 'b'; ADD 'b' AFTER 'a'; ADD 'c' AFTER 'b'` | fails, `22023` |
| `('a', 'd')` | `('a', 'b', 'c', 'd')` | `ADD 'b' AFTER 'a'; ADD 'c' AFTER 'b'` | applies |
| `('b')` | `('a', 'b', 'c')` | `ADD 'a' BEFORE 'b'; ADD 'c' AFTER 'b'` | applies |

So the problematic shape is specifically a contiguous inserted segment at the start of the enum where the segment length is at least two.

## Expected behavior

`pg-delta` should only use an existing label, or a label already added by an earlier statement, as the `BEFORE`/`AFTER` anchor.

For the repro above, one valid order is to add the leading run from right to left:

```sql
ALTER TYPE probe.status ADD VALUE 'b' BEFORE 'c';
ALTER TYPE probe.status ADD VALUE 'a' BEFORE 'b';
```

That order was validated directly on PostgreSQL 16.13 and produces the desired final order `a, b, c`.

## Root cause hypothesis

`diffEnumLabels` simulates additions left-to-right over the branch order. When the first added label is before another added label, `nextBranch` is not in `workingLabels`, but the code still emits `position = { before: nextBranch }`:

```ts
} else if (nextBranch) {
  // nextBranch exists but not in workingLabels yet (shouldn't happen in normal flow)
  position = { before: nextBranch };
  workingLabels.push(newValue);
}
```

That branch is reachable for leading inserted runs such as target labels `a, b, c` from source `c`, and produces `ADD VALUE 'a' BEFORE 'b'` before `b` exists.

## Possible fix direction

When a newly inserted run appears before the first existing label, emit that run from right to left against the nearest existing right-side anchor, or otherwise choose anchors from the current `workingLabels` only.

Focused regression coverage should include:

* failing case: source `('c')`, target `('a', 'b', 'c')`
* failing case: source `('d')`, target `('a', 'b', 'c', 'd')`
* passing controls: one leading value, multiple middle values after an existing value, and leading plus trailing values

pjpjq:
## Summary

Fixes #304.

- schedule pending enum label additions until an adjacent label already exists in the simulated working enum
- emit leading inserted runs from right to left, e.g. `b BEFORE c` then `a BEFORE b`
- keep existing middle/edge addition behavior covered with regression controls

## RED evidence

Before the fix:

- `enum.diff` produced `a BEFORE b` before `b` existed for source `('c')` -> target `('a','b','c')`
- the new PG17 integration regression failed with `22023: "b" is not an existing enum label`

## Validation

- `bun run --cwd packages/pg-delta test src/core/objects/type/enum/enum.diff.test.ts`
- `PGDELTA_SKIP_DUMMY_SECLABEL_BUILD=1 PGDELTA_TEST_POSTGRES_VERSIONS=17 TESTCONTAINERS_RYUK_DISABLED=true bun run --cwd packages/pg-delta test --retry=0 --test-name-pattern "add multiple leading enum values before the first existing value" tests/integration/type-operations.test.ts`
- `PGDELTA_SKIP_DUMMY_SECLABEL_BUILD=1 PGDELTA_TEST_POSTGRES_VERSIONS=17 TESTCONTAINERS_RYUK_DISABLED=true bun run --cwd packages/pg-delta test --retry=0 tests/integration/type-operations.test.ts`
- `bun run format-and-lint`
- `bun run check-types`
- `bun run knip`

Note: I also attempted `PGDELTA_SKIP_DUMMY_SECLABEL_BUILD=1 PGDELTA_TEST_POSTGRES_VERSIONS=17 TESTCONTAINERS_RYUK_DISABLED=true bun run test:pg-delta`; it hit unrelated local Testcontainers/Postgres connection timeouts in non-enum integration files, so I stopped it after confirming the targeted enum coverage and project checks above.

### DIFF
--- a/.changeset/fix-leading-enum-values.md
+++ b/.changeset/fix-leading-enum-values.md
@@ -1,0 +1,5 @@
+---
+"@supabase/pg-delta": patch
+---
+
+Order enum value additions so `BEFORE` and `AFTER` anchors always reference labels that already exist when the statement runs.
--- a/packages/pg-delta/src/core/objects/type/enum/enum.diff.test.ts
+++ b/packages/pg-delta/src/core/objects/type/enum/enum.diff.test.ts
@@ -25,6 +25,20 @@ const testContext = {
   mainRoles: {},
 };
 
+function testEnum(labels: string[]) {
+  return new Enum({
+    schema: "public",
+    name: "e1",
+    owner: "o1",
+    labels: labels.map((label, index) => ({
+      label,
+      sort_order: index + 1,
+    })),
+    comment: null,
+    privileges: [],
+  });
+}
+
 describe.concurrent("enum.diff", () => {
   test("create and drop", () => {
     const props: EnumProps = {
@@ -122,6 +136,30 @@ describe.concurrent("enum.diff", () => {
     expect(add?.position?.after).toBeUndefined();
   });
 
+  test("add multiple leading values from right to left using existing anchors", () => {
+    const main = testEnum(["c"]);
+    const branch = testEnum(["a", "b", "c"]);
+
+    const changes = diffEnums(
+      testContext,
+      { [main.stableId]: main },
+      { [branch.stableId]: branch },
+    );
+    const adds = changes.filter(
+      (c): c is AlterEnumAddValue => c instanceof AlterEnumAddValue,
+    );
+
+    expect(
+      adds.map((add) => ({
+        newValue: add.newValue,
+        position: add.position,
+      })),
+    ).toEqual([
+      { newValue: "b", position: { before: "c" } },
+      { newValue: "a", position: { before: "b" } },
+    ]);
+  });
+
   test("add value in middle (AFTER previous)", () => {
     const main = new Enum({
       schema: "public",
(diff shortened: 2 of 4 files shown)


avallete:
Thanks for working on this! Unfortunately this fix got outpaced by the clean-room engine rewrite (#299, released as `@supabase/pg-delta@1.0.0-alpha.34`): the legacy engine this patches was deleted wholesale, so the branch can no longer be rebased — the files it touches no longer exist.

The underlying bug is fixed in the new engine: the planner anchors leading enum additions on the first *existing* label, and I verified #304's exact repro through the new corpus proof loop. Details on #304.

Closing as superseded — sorry it worked out this way, and thanks again for the contribution. If you'd like to keep contributing, the new engine's corpus (`packages/pg-delta/corpus/`) makes this kind of fix much more approachable.

---

PR #314, opened 2026-06-26
fix(pg-topo): break inline foreign-key cycles instead of dropping statements

### LINKED ISSUE #311: Cycle in schema causing unpredictable breaking diff
# Bug report
- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

I have the following schemas:

```
CREATE TABLE public.note (
  id uuid NOT NULL PRIMARY KEY DEFAULT gen_random_uuid(),
  active_note_version_id uuid NULL REFERENCES public.note_version (id)
);

...

CREATE TABLE public.note_version (
  id uuid NOT NULL PRIMARY KEY DEFAULT gen_random_uuid(),
  content text NOT NULL,
  created_at timestamptz NOT NULL DEFAULT now(),
  note_id uuid NOT NULL REFERENCES public.note (id) ON DELETE CASCADE,
  version_number int DEFAULT 0 NOT NULL,
  title text NOT NULL DEFAULT '',
  previous_note_version_id uuid REFERENCES public.note_version (id)
);
```

For some reason, only when this line is there

```
  active_note_version_id uuid NULL REFERENCES public.note_version (id)
```

The diffing engine seems to end short and drop a bunch of (still present) tables, functions, etc - seeing them as no longer existent. When I comment that line out, everything evaluates normally (no diff).

```
Found drop statements in schema diff. Please double check if these are expected:
DROP FUNCTION private.get_accessible_notes(VARIADIC roles public.access_role[])
DROP FUNCTION public.create_notes(notes jsonb)
DROP TRIGGER increment_note_version ON public.note_version
DROP TRIGGER new_note ON public.note
DROP TRIGGER new_project ON public.project
DROP FUNCTION private.on_new_project()
DROP TRIGGER on_note_organization_share ON public.role_note_organization
DROP TRIGGER on_note_user_share ON public.role_note_user
DROP POLICY "Note-Note link delete policy" ON public.note_note_link
DROP POLICY "Note-Note link insert policy" ON public.note_note_link
DROP POLICY "Note-Note link select policy" ON public.note_note_link
DROP TABLE public.note_note_link
DROP POLICY "Catch-all note_project_save policy" ON public.note_project_save
DROP TABLE public.note_project_save
DROP POLICY "Note version delete policy" ON public.note_version
DROP POLICY "Note version insert policy" ON public.note_version
DROP POLICY "Note version select policy" ON public.note_version
DROP POLICY "Note version update policy" ON public.note_version
DROP TABLE public.note_version
DROP POLICY "Note work link delete policy" ON public.note_work_link
DROP POLICY "Note work link insert policy" ON public.note_work_link
DROP POLICY "Note work link select policy" ON public.note_work_link
DROP TABLE public.note_work_link
DROP POLICY "Note delete policy" ON public.note
DROP POLICY "Note insert policy" ON public.note
DROP POLICY "Note select policy" ON public.note
DROP POLICY "Note update policy" ON public.note
DROP POLICY "note share delete policy" ON public.role_note_organization
DROP POLICY "note share insert policy" ON public.role_note_organization
DROP POLICY "note share select policy" ON public.role_note_organization
DROP POLICY "note share update policy" ON public.role_note_organization
DROP TABLE public.role_note_organization
DROP POLICY "Role note user delete policy" ON public.role_note_user
DROP POLICY "Role note user insert policy" ON public.role_note_user
DROP POLICY "Role note user select policy" ON public.role_note_user
DROP POLICY "Role note user update policy" ON public.role_note_user
DROP FUNCTION private.user_has_note_roles(IN note_id uuid, VARIADIC roles public.access_role[])
DROP TABLE public.role_note_user
DROP TABLE public.note
```

avallete:
## Summary

Fixes supabase/pg-toolbelt#311. A mutual foreign-key reference between tables is a genuine dependency cycle that Kahn's algorithm cannot order, so `analyzeAndSort` previously dropped the cycle participants — and every statement downstream of them — from the `ordered` result, leaving only a `CYCLE_DETECTED` diagnostic. The declarative apply engine then built an incomplete shadow schema from the truncated list, which pg-delta diffed into spurious `DROP` statements for objects that still existed.

The fix mirrors `pg_dump`: when a cycle is detected, cross-cycle FK constraints are deferred into standalone `ALTER TABLE ... ADD CONSTRAINT` statements so the statement set becomes acyclic and applies cleanly.

## Key Changes

* **New module** `src/rewrite/break-fk-cycles.ts`:
  * Identifies `CREATE TABLE` statements participating in each cycle.
  * Strips FK constraints pointing at other tables in the same cycle (self-referential FKs remain inline — PostgreSQL accepts those within a single `CREATE TABLE`).
  * Re-emits each stripped FK as a standalone `ALTER TABLE ... ADD CONSTRAINT`, moving the constraint AST verbatim so the name, columns, and `ON DELETE`/`ON UPDATE` actions are preserved.
  * Column-level FKs (which omit `fk_attrs` because the column is implicit) gain an explicit `fk_attrs` naming the owning column.
  * Handles both table-level and column-level FK constraints, and cycles spanning any number of tables.
* **Updated** `src/analyze-and-sort.ts`:
  * Factored statement-node building into `buildStatementNode` / `buildStatementNodes` for reuse after rewriting.
  * Runs the FK-cycle breaker only when a cycle is present, then re-derives the dependency graph from the rewritten statement set.
  * Safety net: any statement still left in an unbreakable (non-FK) cycle is now appended to `ordered` in deterministic order rather than being dropped, so consumers always receive every statement plus the `CYCLE_DETECTED` diagnostic.
* **Tests + changeset**:
  * `test/cycle-fk-split.test.ts`: statement preservation, cycle resolution, cross-table FK deferral vs. inline self-FK, and ordering.
  * `test/cycle-fk-split.validation.test.ts`: Docker-backed PostgreSQL validation that the rewritten order applies cleanly.
  * Changeset (`@supabase/pg-topo` patch).

## Notes for reviewers

* Synthetic ALTER statements reuse the owning statement's file path with a collision-free `statementIndex`, so the declarative-apply per-input file-path remap keeps working.
* The constraint AST node is moved verbatim, so no FK option is lost.
* The Supabase CLI currently pins `pg-delta@1.0.0-alpha.20`; shipping this to users needs a pg-topo release plus a pg-delta/CLI bump (downstream follow-up).

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

[https://claude.ai/code/session_01DSsCzpn9AZC52rUj8tmjHz](<https://claude.ai/code/session_01DSsCzpn9AZC52rUj8tmjHz>)

### DIFF
--- a/.changeset/break-inline-fk-cycles.md
+++ b/.changeset/break-inline-fk-cycles.md
@@ -1,0 +1,22 @@
+---
+"@supabase/pg-topo": patch
+---
+
+fix(pg-topo): break inline foreign-key cycles instead of silently dropping statements
+
+A mutual foreign-key reference between two tables (or any FK loop spanning
+several tables) is a real dependency cycle: neither table can be created with
+its FK declared inline. Previously the topological sort dropped the cycle
+participants — and every statement depending on them — from the `ordered`
+result, leaving only a `CYCLE_DETECTED` diagnostic. Declarative apply then
+built an incomplete schema from the truncated list, which surfaced downstream
+as spurious `DROP` statements for objects that still existed
+(supabase/pg-toolbelt#311).
+
+`analyzeAndSort` now mirrors `pg_dump`: when a cycle is detected it strips the
+cross-cycle FK constraints from the offending `CREATE TABLE` statements and
+re-emits them as standalone `ALTER TABLE ... ADD CONSTRAINT` statements,
+preserving the constraint name, columns, and `ON DELETE`/`ON UPDATE` actions.
+Self-referential FKs are kept inline. As a safety net, any statement still
+left in an unbreakable cycle is appended to `ordered` in deterministic order
+rather than being dropped, so consumers always receive every statement.
--- a/packages/pg-topo/src/analyze-and-sort.ts
+++ b/packages/pg-topo/src/analyze-and-sort.ts
@@ -16,6 +16,7 @@ import type {
   GraphReport,
   StatementNode,
 } from "./model/types.ts";
+import { breakForeignKeyCycles } from "./rewrite/break-fk-cycles.ts";
 
 const dedupeDiagnostics = (diagnostics: Diagnostic[]): Diagnostic[] => {
   const map = new Map<string, Diagnostic>();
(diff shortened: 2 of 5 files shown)


avallete:
Will be fixed by new cutover #299

PR #329, opened 2026-07-10
fix(pg-delta): make schema apply --profile supabase work for non-superuser Cloud users

avallete:
## What & why

Makes `pgdelta schema apply --profile supabase` work end-to-end for a **real, non-superuser Supabase Cloud user** (the `postgres` role Cloud actually hands out — a non-superuser member of `supabase_privileged_role`, not the `supabase_admin` the test harness previously stood in with).

The Supabase roundtrip acceptance harness (`scripts/roundtrip-supabase.ts`) was run against a real Cloud project as that faithful role. The prior `SUPERUSER` test shim had been **masking a series of real non-superuser failures**; each is fixed here RED-first. Apply now fully succeeds (2746 actions) against a schema-rich real project, with the roundtrip diff down to one documented baseline-drift line.

## Units (each a RED→GREEN pair)

- **A — composite attribute order** (`extract/types.ts`, `plan/rules/types.ts`): carry `attnum` as non-semantic `_position`; render `CREATE TYPE` in declared order so reconstruction can't change row layout. (Codex `3537910546`, render half.)
- **fixture** — regenerate `tests/fixtures/supabase-base-init/17.sql` (was baked pre-A).
- **B — body-validation scoping + diagnostics** (`load-sql-files.ts`, `cli/main.ts`): seeded-schema routine-body failures become named warnings (not hard errors); user-schema failures still block and name the routine; CLI prints `ShadowLoadError.details`.
- **C — non-superuser-replayable seed** (`seed-assumed-schemas.ts`): omit `defaultPrivilege` facts and skip routines whose `pg_proc.proconfig` sets a superuser-only GUC (carried structurally as `_configGucs`; **no SQL-text editing**) plus in-seed dependents.
- **D — constraints on deferred columns** (`plan/internal.ts`): stop folding a table constraint inline when a same-table column it references was deferred to a later `ADD COLUMN` (domain-typed / generated) — render a standalone `ADD CONSTRAINT`.
- **E — system-role ADP** (`policy/supabase.ts`): exclude `defaultPrivilege` entries declared FOR a system role (platform-managed); grantee-side kept.
- **G — validation language scope** (`load-sql-files.ts`): validate only `sql`/`plpgsql` bodies (what `check_function_bodies` checks); skip `internal`/`c` (e.g. `CREATE TYPE … AS RANGE` support functions).
- **H — resolve policy view before scope projection** (`plan/phases/change-set.ts`, `plan.ts`, `apply.ts`, `prove.ts`, `cli/commands/schema.ts`): thread management scope into the planner so the managed view is `projectManagementScope(resolveView(raw, …), scope)` in one place. Owner edges survive policy resolution, so owner-based exclusion (Supabase Rule 6) correctly excludes system-role-owned platform objects instead of planning a `DROP` the applier can't run. (This had been silently dropping platform event triggers under the shim.)
- **harness** — `tests/containers.ts` provisions `postgres` as the faithful non-superuser privileged role.
- **docs** — `docs/roadmap/pg-delta-next-follow-ups.md` records the fixes + the remaining baseline-drift follow-up.

## Verification

- Corpus (postgres:17-alpine): **592/592**
- Unit suite: **685/685**
- Supabase-gated e2e: **9/9**
- Acceptance harness: apply succeeds (2746 actions); roundtrip residue = only `schemas/public/default_privileges.sql` (local-`supabase start`-vs-Cloud baseline drift — versioned-baseline-sidecar backlog, not engine-fixable).

## Not in this PR

Unit F (pg-topo cyclic-bundled-statement decomposition + `--dump-applied-sql`) is deferred to a follow-up.


packages/pg-delta/src/cli/commands/schema.ts
```diff
@@ -806,6 +806,17 @@ export async function cmdSchemaApply(args: string[]): Promise<void> {
...
+        // privileged NON-superuser, so a seeded routine's `SET <suser-guc> TO …`
+        // header clause (e.g. realtime.list_changes' `SET log_min_messages`)
+        // fails to REPLAY (42501). The clause is never semantically compared (the
+        // seeded routine re-extracts reference-only and cancels), so it is safe to
+        // strip from the seed DDL. The target shares the co-located shadow's
+        // cluster, so the target's GUC catalog is authoritative for the shadow.
+        const susetRows = await tgt.pool.query<{ name: string }>(
+          `SELECT name FROM pg_settings WHERE context = 'superuser'`,
+        );
+        const susetGucs = new Set(susetRows.rows.map((r) => r.name));
```
avallete: **issue**

I think it's bad that supabase scope leak into the general cli here. That needs more thinking.

packages/pg-delta/src/extract/roles.ts
```diff
@@ -111,12 +127,16 @@ export async function extractRolesAndGrants(
...
+      -- grantees that HAVE a built-in default but are ABSENT (revoked) — except
+      -- the grantor's own self-entry, whose absence is a behavioral no-op
+      -- (Postgres re-adds the owner's acldefault entry at object-creation time),
+      -- so canonicalize by never emitting a marker for it.
       SELECT CASE WHEN dd.grantee_oid = 0 THEN 'PUBLIC'
                   ELSE (SELECT rolname FROM pg_roles WHERE oid = dd.grantee_oid) END,
              ARRAY[]::text[], NULL::text[], dd.privileges
       FROM def dd
       WHERE NOT EXISTS (SELECT 1 FROM stored s WHERE s.grantee_oid = dd.grantee_oid)
+        AND dd.grantee_oid <> d.defaclrole
```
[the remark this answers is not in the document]
avallete: Worth double checking, this needs to be asserted in a test.

packages/pg-delta/src/frontends/export-sql-files.ts
```diff
@@ -567,16 +567,6 @@ function pathFor(id: StableId, ctx: PathContext): string {
...
-    if (kind === "trigger")
-      console.error(
-        "DBG trigger",
-        JSON.stringify({
-          key: `${t.schema} ${t.table}`,
-          hit: ctx.relationDir.get(nameKey(t.schema, t.table)),
-          mapSize: ctx.relationDir.size,
-          keys: [...ctx.relationDir.keys()],
-        }),
-      );
```
avallete: **question**

Why is that deleted ?

packages/pg-delta/src/frontends/seed-assumed-schemas.ts
```diff
@@ -30,8 +30,29 @@
...
+ *     via `susetGucs` ∩ the routine's structured `_configGucs` (from
+ *     `pg_proc.proconfig`) — never by parsing the `def`. The one real occurrence
+ *     on the Supabase surface is `realtime.list_changes`.
+ * A seeded fact is reference-only on both sides, so its absence is symmetric and
+ * cancels in the diff. If a user file genuinely references a skipped routine the
+ * load fails LOUDLY at file:line with a precise missing-object error — acceptable
+ * (user code essentially never calls these internal platform RPCs, and a clear
+ * error beats rewritten SQL). `susetGucs` absent ⇒ nothing is skipped for this
+ * reason (byte-identical behavior). Any fact that DEPENDS on a skipped routine is
+ * transitively skipped too (it could not replay against a missing dependency).
```
avallete, line 53: **question**

Should that only occurs when non-superuser is used ?

packages/pg-delta/src/extract/roles.ts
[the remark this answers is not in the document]
avallete: Confirmed, with a nuance: the literal minimal repro (bare global self-revoke, nothing else granted) is actually still a no-op — the created object's relacl degenerates to NULL and the owner keeps its privileges (verified on PG 17). But the general claim holds: on a global row that still carries other grantees, Postgres applies the stored ACL verbatim, so the owner self-revoke is real and was being dropped.

Fixed in 7f68786b: suppression is now conditional (per-schema and bare-empty-global stay canonicalized no-ops), with a focused integration test covering all three shapes and a corpus scenario locking the round-trip both directions (47092f30).

packages/pg-delta/src/frontends/seed-assumed-schemas.ts
```diff
@@ -88,9 +129,60 @@ export function deriveAssumedSchemaSeed(
...
+            changed = true;
+          }
+        }
+      }
+    }
+  }
+
+  const seedFacts =
+    excluded.size > 0
+      ? keptFacts.filter((f) => !excluded.has(encodeId(f.id)))
```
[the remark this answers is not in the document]
avallete: Confirmed — containment is `Fact.parent`, not a `depends` edge, so an excluded container orphaned its children and `buildFactBase` threw. Dormant on today's Supabase surface (only `realtime.list_changes` is excluded and nothing with children depends on it), but fixed anyway in f60b06d0: exclusion now cascades over depends edges and parent chains to a joint fixpoint (joint because pg_depend endpoints can resolve to column-level ids). Unit test with a container+child dependent in 4730aa91.

packages/pg-delta/src/cli/commands/schema.ts
avallete: Dug into this — the mechanism is actually profile-generic (gated on `Policy.assumedSchemas`, no supabase coupling anywhere in the seed/load path; any custom profile declaring assumed schemas hits it). The real smells are the hand-rolled catalog query living inline in `cmdSchemaApply`, and that it isn't gated on the applier being a non-superuser.

Plan (follow-up PR): move the `pg_settings` probe into `resolveProfile` next to the existing capability probe (which already checks `rolsuper` on the same pool), gate it on `assumedSchemas` non-empty AND `!isSuperuser`; `deriveAssumedSchemaSeed` stays pure/sync taking a `susetGucs` set so its Docker-free unit tests survive.

packages/pg-delta/src/frontends/export-sql-files.ts
avallete: Leftover `console.error` debug logging that accidentally shipped with the #323 export-as-source-of-truth work — this just removes it. No behavior change.

packages/pg-delta/src/frontends/seed-assumed-schemas.ts
avallete: Today it fires whenever the profile declares assumed schemas, regardless of role. It's semantically safe either way (seeded facts are reference-only and cancel symmetrically in the diff), but yes — gating on `!isSuperuser` is the better shape, and it falls out of the same `resolveProfile` refactor planned in the thread on `cmdSchemaApply`.

packages/pg-delta/src/cli/commands/schema.ts
avallete: Update: landed in this PR instead of a follow-up. ca6cf5fb moves the probe into `resolveProfile` as `ResolvedProfile.susetGucs` (generic profile-runtime context, gated on `Policy.assumedSchemas`), and it's now also gated on the applier actually being a non-superuser — a superuser applier seeds SUSET-proconfig routines instead of skipping them. RED-first coverage in `tests/resolve-profile-suset.test.ts` (c304237b).

avallete:
@codex review

packages/pg-delta/src/integrations/profile.ts
```diff
@@ -130,6 +138,34 @@ export async function resolveProfile(
...
+  // when the applier is NOT a superuser (a superuser needs no stripping — the
+  // seed's routine replays as-is). Reuses the capability probed above when
+  // `restrictToApplier` was requested; otherwise probes locally without
+  // threading that probe into `planOptions`/`capability` (those stay governed
+  // strictly by `restrictToApplier`). The `pool` this resolves against is the
+  // same connection `schema apply` extracts the target from, which shares the
+  // co-located shadow's cluster + role, so its GUC catalog and role are
+  // authoritative for the shadow.
+  let susetGucs: ReadonlySet<string> | undefined;
+  if ((policy?.assumedSchemas?.length ?? 0) > 0) {
```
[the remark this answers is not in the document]
avallete: Confirmed — `flattenPolicy` aggregates `assumedSchemas` across the `extends` chain and `cmdSchemaApply` seeds from the flattened set, so the probe gate reading only the policy's own field was inconsistent. Fixed in 3b6ffb61: `resolveProfile` now gates on the flattened policy's assumed schemas. RED-first coverage with an `extends`-inherited policy in 0dcd0d9e.

packages/pg-delta/src/frontends/load-sql-files.ts
```diff
@@ -707,26 +710,58 @@ export async function loadSqlFiles(
...
-          message: error instanceof Error ? error.message : String(error),
-        });
+        const message = `${row.nspname}.${row.proname}: ${error instanceof Error ? error.message : String(error)}`;
+        // Seeded platform routines (Phase 2b assumed schemas) are reference-only
+        // on both sides of the diff — they cancel, so a wonky seeded body cannot
+        // corrupt the plan — and they are not the user's code to fail their
+        // apply on. Still surface it as a warning rather than swallowing it: a
+        // seeded-routine validation failure has previously exposed a real
+        // engine bug in platform-code reconstruction.
+        if (seededSchemas.includes(row.nspname)) {
```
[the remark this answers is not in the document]
avallete: Confirmed — and it's actually worse than framed: a user routine in an assumed schema is also reference-only in the diff, so beyond the soft warning it was never planned at all (warn + silent no-op on the target; `--strict-coverage` doesn't catch it either, since the kind is modeled and no `unmodeled_kind` diagnostic fires).

Fixed in ca81ecbb (RED 15744b9f): the seed now returns the overload-safe encoded identities of the routines it actually created (mapped to their seeded defs), and the leniency gate requires identity membership AND a byte-identical def — so a new user routine, a new overload of a seeded name, and a `CREATE OR REPLACE` of a seeded body all fail loudly. Verified against the real Supabase seed with zero spurious throws on genuine platform routines.

packages/pg-delta/src/extract/types.ts
```diff
@@ -182,7 +190,17 @@ export async function extractTypes(ctx: ExtractContext): Promise<void> {
...
+        // `_position` is the declared attribute position (pg_attribute.attnum).
+        // Row layout — hence render order — is desired state, but positional
+        // identity is not (attributes are name-keyed sub-facts, like columns),
+        // so the `_`-prefix excludes it from the hash and diff (core/hash.ts,
+        // core/diff.ts) while the composite CREATE TYPE rule still renders
+        // attributes in this order (plan/rules/types.ts).
+        payload: {
+          type: a.type,
+          collation: a.collation ?? null,
+          _position: a.position,
```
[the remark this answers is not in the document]
avallete: Valid, but deliberately out of scope for this PR — this is the deferred diff-half of Unit A (the PR fixed render fidelity only), and it's the same systemic class as your earlier table-column-order finding: physical order isn't semantic state anywhere in the fact base. Fixing it properly means making `_position` semantic (which changes fact hashes for every composite type and invalidates all captured baselines — committed policy baselines, fixtures, users' snapshot artifacts) and building type-replace + dependent-rebuild machinery, since no ALTER exists for in-place reorder.

Tracked as #330 with the design options (semantic position + replace machinery vs fail-loudly-on-reorder vs hybrid), covering both composites and table columns.

Issue #333, opened 2026-07-16
pg-delta: plan/apply correctness gaps (review-backlog triage from #299)


avallete:
Tracking issue for **plan/apply correctness gaps** in the `pg-delta` clean-room rewrite, surfaced by the accumulated automated-review backlog on supabase/pg-toolbelt#299 and triaged (0 false positives in this batch). These are genuine bugs (apply/plan/extract failures), distinct from the fidelity follow-ups tracked in supabase/pg-toolbelt#332. Ordered roughly by severity.

### 🔴 Destructive / P1

1. `DROP OWNED BY` **on role removal can destroy unmanaged objects.** `plan/rules/roles.ts` role drop emits `DROP OWNED BY x; DROP ROLE x`; if the dropped role owns anything outside the managed/projected view, apply deletes real user data. Prefer targeted revoke/reassign, or fail loud on residual ownership.
2. **In-place alters don't participate in the dependency walk (ordering family).** `plan/internal.ts` order walk only follows `action.produces`; alter specs neither consume their new targets nor release `from`-side refs. Four manifestations, all apply-failures:
   * `ALTER COLUMN … TYPE <new enum/domain>` sorts before the `CREATE TYPE` (`plan/rules/tables.ts` has no `consumes`).
   * `ALTER EXTENSION … SET SCHEMA new` vs `DROP SCHEMA old` (`plan/rules/schemas.ts`).
   * sequence `OWNED BY new` vs `DROP TABLE old` (`plan/rules/helpers.ts` `sequenceOwnedBySpecs`).
   * `ALTER POLICY … TO` releasing a role the same plan drops (`plan/rules/policies.ts`).
3. **Replacing a foreign server with existing children** — `DROP SERVER` (RESTRICT) fails while foreign tables/user-mappings still depend on it; `foreignTable` isn't a rebuildable kind. *(UNSURE — needs a repro to confirm before fixing.)*
4. **User table rewrite-rule deps unresolved** — `extract/dependencies.ts` `rw` CTE only maps `relkind IN ('v','m')`, so a rule on a plain table referencing a function isn't rebuilt before `DROP FUNCTION` → apply fails.

### 🟠 Apply/plan failures

 5. **Domain replace with in-use columns has no guard** — `plan/rules/types.ts` `baseType`/`collation: "replace"` emits `DROP DOMAIN` that PG rejects while columns use it (range types already have this guard; domains don't).
[...]

Issue #332, opened 2026-07-16
pg-delta: extraction/model fidelity gaps (post-alpha follow-ups from #299 review)


avallete:
Tracking issue for extraction/model fidelity gaps surfaced by review on the clean-room rewrite (supabase/pg-toolbelt#299). These are real but non-blocking for the `BREAKING alpha` — schemas differing only by these attributes currently hash equal / a from-empty export can lose or mis-set them (no apply crash). Triaged and deferred here; the apply-failure and cheap/high-value items were fixed in supabase/pg-toolbelt#299 directly.

### 1. Composite type attribute order not diffed

Tracked in depth in supabase/pg-toolbelt#330, which frames this as a design question with three options (semantic position plus replace machinery, fail loudly on reorder, or hybrid) and covers the table-column half as well. This entry stays as the review record; the work happens there.

Composite attribute position is stored under `_position`, which `canonicalize`/`hash`/`diff` deliberately exclude (`src/extract/types.ts`, `src/core/hash.ts`, `src/core/diff.ts`). From-empty export IS order-correct (create sorts by `_position`), but an *incremental* attribute reorder / mid-insert on an existing composite type is invisible and the proof falsely converges. Fix needs an order-aware sibling check that seeds a type replace (reusing the enum-style column-migration path) — a design decision, not a naive hash change (naive hashing would force destructive rebuilds on every reorder and break name-keyed rename detection).

### 2. Column STORAGE / COMPRESSION / per-column FDW options not extracted

`pg_attribute.attstorage` / `attcompression` / `attfdwoptions` are never read (`src/extract/relations.ts`); a schema differing only by `SET STORAGE`/`SET COMPRESSION` (or a foreign-table column's `OPTIONS`) hashes equal and a from-empty export recreates with type defaults. Fix needs pg_dump-style default suppression (`attstorage <> typstorage`) + a PG13 gate for `attcompression` + column-attribute rules (`SET STORAGE`/`SET COMPRESSION`/`ALTER COLUMN OPTIONS`).

### 3. Multiple inheritance parents (full support)

[...]

PR #349, opened 2026-07-20
fix(pg-delta): avoid role rename policy dependency cycle

avallete:
## What kind of change does this PR introduce?

Bug fix for the B1 crash-class planner issue: an accepted role rename combined with an RLS policy referencing that role produced a dependency cycle. Includes a patch changeset for `@supabase/pg-delta`.

## What is the current behavior?

With `renames: "auto"`, changing `role_a` to `role_b` while an RLS policy changes its `TO` payload from `role_a` to `role_b` creates contradictory graph edges:

- consuming `role_b` orders `ALTER ROLE ... RENAME` before `ALTER POLICY ... TO role_b`;
- releasing `role_a` orders the policy alteration before the same rename.

Planning then throws:

```text
dependency cycle among 2 actions:
  ALTER ROLE "role_a" RENAME TO "role_b"
  ALTER POLICY "docs_read" ON "app"."docs" TO "role_b"
```

Refs #333

## What is the new behavior?

The action graph skips the conflicting release-before-destroy edge only when the destroyer is an accepted role rename that destroys the released old role and produces the new role consumed by the releasing action. Genuine role drops retain their existing release ordering.

Coverage includes the in-memory RED regression, a negative over-skip test, and an isolated-cluster extract → plan → prove integration test.

## Testing

- RED confirmed in both unit and integration tests with the cycle above.
- `bun run test:pg-delta` — 790 passed.
- `PGDELTA_TEST_IMAGE=postgres:17-alpine bun test tests/renames.test.ts` — 12 passed.
- `PGDELTA_NEXT_PROGRESS=1 PGDELTA_TEST_IMAGE=postgres:17-alpine bun test tests/engine.test.ts` — 632 passed.
- `bun run format-and-lint`, `bun run check-types`, and `bun run knip` — passed.
- CI: unit, integration, and corpus matrices across PostgreSQL 14–18 are green.

## Additional context

This is intentionally a narrow ordering carve-out. The planned I1 pre-diff role-reference normalization will make the policy mutation redundant and remove this temporary exception.

### DIFF
--- a/.changeset/famous-eagles-fry.md
+++ b/.changeset/famous-eagles-fry.md
@@ -1,0 +1,5 @@
+---
+"@supabase/pg-delta": patch
+---
+
+Fix dependency-cycle planning when an accepted role rename is referenced by an RLS policy.
--- a/packages/pg-delta/src/plan/internal.ts
+++ b/packages/pg-delta/src/plan/internal.ts
@@ -123,10 +123,40 @@ export function buildActionGraph(
     alterersOf.set(key, list);
   });
 
+  // B1 temporary carve-out: an accepted ROLE rename destroys the old role id
+  // and produces the new one in the SAME action. A payload mutation such as
+  // ALTER POLICY … TO releases the old id and consumes that exact new id. Its
+  // consume edge correctly orders rename -> mutation; adding the generic
+  // release-before-destroy edge in the opposite direction would make a cycle.
+  // Keep this deliberately narrower than "release targets a rename": require
+  // one accepted role old->new pair represented by the destroyer action AND
+  // require this action to consume that pair's new id. I1's pre-diff payload
+  // normalization makes the redundant policy mutation disappear and removes
+  // this carve-out.
+  const releasesAcceptedRoleRename = (
+    action: Action,
+    released: StableId,
+    destroyer: number,
+  ): boolean => {
+    if (released.kind !== "role" || !renameActionIndices.has(destroyer)) {
+      return false;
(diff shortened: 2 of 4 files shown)


avallete:
Thanks for the audit. I fact-checked all nine findings against the PR diff, current planner/extractor code, corpus coverage, and the existing roadmap.

PR #349 changes only four files: the B1 changeset, `plan/internal.ts`, and the two rename regression test files. None of the reported lines is changed by this PR, so none is a B1 merge blocker.

Assessment:

- Eight findings identify real pre-existing fidelity/safety gaps: membership `INHERIT`/`SET`, subscription `password_required`/`failover`, role connection limits, materialized-view population state, expression-index statistics targets, ICU collation `RULES`, unlogged sequence persistence, and non-relation coverage in the reused-shadow emptiness guard.
- Most are already recorded in `docs/roadmap/pg-delta-next-follow-ups.md`; the two missing details (expression-index statistics and subscription `password_required`) are now recorded on #332.
- The foreign-server child-rebuild finding is stale/invalid. Replacement emission recursively drops non-cascading children and recreates surviving descendants (`plan/phases/action-emitter.ts`), and the exact server + foreign table + user mapping case is proved in both directions by `foreign-data-wrapper-operations--replace-server-with-children`. That fix landed in `da2d75c1`, before this PR's base.

No B1 code change is warranted from this review. The valid items remain follow-up work under #332/#333 so this urgent crash fix stays one-track/one-PR.

PR #351, opened 2026-07-21
refactor(pg-delta): centralize managed view reconstruction

avallete:
## Summary
- add an internal `reconstructManagedView` helper that fixes the `resolveView` → `projectManagementScope` order
- migrate plan, prove, apply, and schema export to the shared reconstruction path
- add source guards and behavioral pins, including a reverse-order regression case and full FactBase comparison
- document the internal reconstruction seam; no changeset because behavior and public API are unchanged

## Test plan
- [x] `bun test src/policy/reconstruct.test.ts`
- [x] `bun test src/policy/ src/plan/phases/ src/proof/prove.test.ts`
- [x] `bun test tests/schema-frontends.test.ts`
- [x] `bun run format-and-lint`
- [x] `bun run check-types`

### DIFF
--- a/docs/architecture/managed-view-architecture.md
+++ b/docs/architecture/managed-view-architecture.md
@@ -222,6 +222,12 @@ function plan(rawSource, rawDesired, options) {
   edges). The Supabase scope rules are more cases. One code path.
 - The proof loop’s `reextract` is wrapped in the same `resolveView` so
   `plan == prove == run` holds with no special-casing.
+- **Reconstruction under management scope** (plan / prove / apply / schema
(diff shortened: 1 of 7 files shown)


avallete (commented):
[P3] Preserve warning callback ordering in packages/pg-delta/src/frontends/schema-export.ts:104-114.

Previously resolveView ran before onWarning; reconstruction now runs afterward. Because resolveProfile retains the caller-owned mutable policy reference, a warning callback can mutate policy filters and change the exported schema. Move warning delivery after reconstructManagedView to preserve the prior behavior.

PR #414, opened 2026-08-13
feat(pg-delta): export a pure SQL-file classification helper

avallete:
## Summary

- First slice of [schema-first CLI enablement](docs/roadmap/schema-first-cli-enablement.md) (WP3a): export `classifySqlFiles` / `classifySqlContent` so the Supabase CLI can compose `schema pull` without reimplementing created/updated/unchanged/removed/unmanaged.
- Classification is **pure** (in-memory proposed files vs existing tree vs the previous export's owned-file list). `_custom/` stays invisible, matching the exporter reservation.
- `writeExportFiles` is **not** exported — staging, unmanaged-file refusal, `_custom/README.md` scaffolding, and install stay with `pgdelta` / the CLI adapter. `schema export` now uses `classifySqlContent` for the byte-compare path.

## Test plan

- [x] RED: `bun test src/frontends/classify-sql-files.test.ts` failed with `Cannot find module './classify-sql-files.ts'`
- [x] GREEN: 14 classification unit tests pass (created/updated/unchanged, removed vs unmanaged, `_custom/` reservation, separator normalization)
- [x] Existing `writeExportFiles` cases in `collect-sql-files.test.ts` still pass
- [x] `public-api.test.ts` pins the root re-export
- [x] `bun run format-and-lint && bun run check-types`

Follow-ups (not this PR): WP3b re-exports (`pruneStaleSqlFiles`, `renderApplyScript`, `probeUnmodeledIdentitiesPinned`), then WP3c segments, WP1 `planId`, WP2 hazards.

### DIFF
--- a/packages/pg-delta/src/frontends/classify-sql-files.ts
+++ b/packages/pg-delta/src/frontends/classify-sql-files.ts
@@ -24,7 +24,8 @@ export interface ClassifySqlFilesInput {
    *  are POSIX-relative. */
   proposed: readonly SqlFile[];
   /** Existing `.sql` file bodies keyed by path relative to the export root.
-   *  Either separator is accepted. Non-`.sql` keys and `_custom/**` are ignored. */
+   *  Either separator is accepted; `.` segments are dropped. Non-`.sql` keys
+   *  and `_custom/**` are ignored. `..` is caller-owned (left as-is). */
   existing: ReadonlyMap<string, string>;
   /** POSIX-relative paths the previous export owned (`ExportManifest.files`).
    *  Absent (pre-feature / hand-authored dir) → every extra `.sql` is
@@ -43,11 +44,14 @@ export interface SqlFileClassification {
   unmanaged: string[];
 }
 
-/** Normalize a relative path to POSIX (`/` separators, no empty segments). */
+/** Normalize a relative path to POSIX (`/` separators, no empty or `.`
+ *  segments). `..` is left as-is — callers must not pass parent-directory
+ *  segments; this helper is pure, so `..` is a mismatch hazard, not a
+ *  traversal one. */
 function posixRelPath(relPath: string): string {
   return relPath
     .split(/[\\/]/)
-    .filter((segment) => segment !== "")
+    .filter((segment) => segment !== "" && segment !== ".")
     .join("/");
 }
 
--- a/packages/pg-delta/src/frontends/classify-sql-files.ts
+++ b/packages/pg-delta/src/frontends/classify-sql-files.ts
@@ -1,0 +1,119 @@
+/**
+ * Pure export-file classification for declarative schema trees.
+ *
+ * Compares a proposed `SqlFile[]` (what an export would write) against an
+ * in-memory existing tree and the previous export's owned-file list
+ * (`.pgdelta-export.json` `files`). Returns created / updated / unchanged /
+ * removed / unmanaged as POSIX-relative paths. No filesystem mutation: the
+ * caller owns staging, `--force` / `--prune-unmanaged` authorization, and
+ * install (schema-first CLI enablement WP3a; the RFC's "reusable typed file
+ * classification helper").
+ *
+ * `_custom/` is invisible here, matching the exporter reservation
+ * (`custom-dir.ts`): nothing under the root-level folder is created, updated,
+ * unchanged, removed, or unmanaged. Nested `schemas/app/_custom/` is ordinary
+ * managed space.
+ */
+import type { SqlFile } from "./load-sql-files.ts";
+import { isCustomPath } from "./custom-dir.ts";
+
+export type SqlFileChange = "created" | "updated" | "unchanged";
+
+export interface ClassifySqlFilesInput {
+  /** Files the export would write. Names may use either separator; results
+   *  are POSIX-relative. */
+  proposed: readonly SqlFile[];
+  /** Existing `.sql` file bodies keyed by path relative to the export root.
+   *  Either separator is accepted. Non-`.sql` keys and `_custom/**` are ignored. */
+  existing: ReadonlyMap<string, string>;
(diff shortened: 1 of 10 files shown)


packages/pg-delta/src/frontends/classify-sql-files.ts
```diff
@@ -0,0 +1,119 @@
...
+function posixRelPath(relPath: string): string {
+  return relPath
+    .split(/[\\/]/)
+    .filter((segment) => segment !== "")
+    .join("/");
+}
+
+function isManagedSqlPath(path: string): boolean {
+  return path.endsWith(".sql") && !isCustomPath(path);
+}
```
avallete: **SIGNIFICANT — `.` segments survive normalization, and the miss also punctures the `_custom/` reservation**

This confirms the Codex comment above — I reproduced it at this commit, and it's slightly worse than stated:

- `existing` key `./schemas/app/t.sql` vs proposed `schemas/app/t.sql` (identical bytes) → the same file is reported both `created` **and** `unmanaged`:
  ```json
  {"created":["schemas/app/t.sql"],"unmanaged":["./schemas/app/t.sql"], ...}
  ```
- `existing` key `./_custom/x.sql` → reported `unmanaged`, because `isCustomPath` matches the first segment and sees `"."`, not `"_custom"`. That violates this module's own documented invariant ("nothing under the root-level folder is … removed, or unmanaged") — the guarantee that keeps a schema-pull consumer from refusing on the escape hatch.

Since this is a new public API whose whole purpose is composition by an external CLI, and the doc comment promises normalization ("Either separator is accepted", "no empty segments"), a `./`-prefixed key from a directory walk is well within the input contract.

Smallest fix: add `segment !== "."` to the filter here; decide `..` explicitly — either resolve it with a segment stack or document it as caller responsibility (the helper is pure, so `..` is a mismatch hazard, not a traversal one). Two tests to pin it: a `./`-prefixed key matching its plain spelling, and `./_custom/x.sql` staying invisible.

---

avallete (commented):
## Overall assessment

The implementation is sound and well-scoped. The helper is genuinely pure, and its `removed` / `unmanaged` / `_custom/` semantics exactly mirror the pruner's contract (`pruneStaleSqlFiles`: `previouslyOwned` → removed, else unmanaged, absent manifest → all unmanaged, root `_custom/` invisible) — so there's one decision table, not two. The `writeExportFiles` rewiring is behavior-preserving: the size fast-path, buffered-read-only-on-plausible-match, and the ENOENT-only "created" rule (PR #405 review) all survive intact. Deliberately *not* exporting `writeExportFiles` keeps staging/refusal/scaffolding policy with the CLI, which is the right boundary.

Verified at `fa3bcb9`: the 14 new classification tests plus `public-api.test.ts` pass (21 pass, 0 fail). Changeset, roadmap docs, and the RED step are all in order.

One real defect (inline) — the Codex P2 is valid and slightly understated.

## Findings

- **SIGNIFICANT** (inline on `posixRelPath`): `.` segments survive normalization, so `./`-prefixed keys misclassify as `created` + `unmanaged`, and `./_custom/x.sql` escapes the `_custom/` reservation entirely. One-line fix plus two tests.
- **NON-BLOCKING**: names that collide after normalization (e.g. `a//b.sql` vs `a/b.sql`) silently last-write-win for content while keeping first-occurrence order. No realistic producer emits collisions — a one-line note in the `ClassifySqlFilesInput` doc would do, if you're touching the file anyway.

## Verdict

**Changes requested** (submitted as a comment review — GitHub does not allow request-changes on one's own PR): one small, well-understood fix away from approval. Everything else is approve-quality.

---

packages/pg-delta/src/frontends/classify-sql-files.ts
```diff
@@ -0,0 +1,119 @@
...
+   *  `proposed` does not claim. */
+  unmanaged: string[];
+}
+
+/** Normalize a relative path to POSIX (`/` separators, no empty segments). */
+function posixRelPath(relPath: string): string {
+  return relPath
+    .split(/[\\/]/)
+    .filter((segment) => segment !== "")
+    .join("/");
```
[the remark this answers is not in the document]
avallete: Fixed in 8dfa2b6d. `posixRelPath` now drops `.` segments the same way as empty ones, so `./schemas/app/t.sql` classifies as unchanged against `schemas/app/t.sql` instead of created+unmanaged. `..` is left as-is and documented as caller-owned (this helper is pure, so it is a mismatch hazard rather than a traversal one).

packages/pg-delta/src/frontends/classify-sql-files.ts
avallete: Fixed in 8dfa2b6d. Dropping `.` segments means `./_custom/x.sql` normalizes to `_custom/x.sql` before `isCustomPath`, so the reserved folder stays invisible. Tests cover the `./`-prefixed match and the `./_custom/` invariant.

avallete (commented):
## Round 2 — fix verified

Re-reviewed at `8dfa2b6`. The dot-segment fix is exactly the smallest change asked for, and I verified it empirically:

- `./schemas/app/t.sql` vs `schemas/app/t.sql` (identical bytes) → now `unchanged` (was `created` + `unmanaged`).
- `./_custom/x.sql` → now invisible (was `unmanaged`), restoring the reserved-folder invariant before `isCustomPath` runs.
- `..` left as-is and documented as caller-owned on both `posixRelPath` and the `existing` field — the right call for a pure helper (mismatch hazard, not traversal).
- RED/GREEN split checks out: at `41a902f` the two new tests fail (2 fail / 14 pass); at `8dfa2b6` all 23 tests pass (classification + public-api).

No new findings. The earlier non-blocking note (documenting last-wins for names that collide after normalization) remains optional — fine to leave.

## Verdict

**Approve** (as a comment review — GitHub does not allow approving one's own PR from this token). Good to merge once CI is green.

---

jgoux (approved):

PR #475, opened 2026-09-13
fix(pg-delta): round-trip overlay default grants on auto-expose baselines

avallete:
## Summary

`load(export(live))` onto a Supabase auto-expose-ON baseline now preserves live ACLs. Absence of `anon` (or another overlay grantee) is a create-time `REVOKE ALL`, not a silent keep of the load environment’s injectee. A desired default privilege that is a subset of the dest injectee is wipe-then-grant (`REVOKE ALL` then `GRANT` the live set). Overlay `default_privileges.sql` loads before object files so dest auto-expose does not inject grants onto unmodeled identity sequences.

Overlay tuples live on the policy (`assumedDefaultGrants`: creating role, schema, objtype, grantee — no privilege lists). The planner still has no platform role names of its own. Empty-fact ADP wipes stay export-options-only so policy-only DB-to-DB plans are unchanged.

Also on this branch (unrelated to privileges): corpus `dependencies-cycles--partitioned-parent-publication-via-root` (df50fe4) locks a partitioned parent + `publish_via_partition_root` replace path.

## Linked issue

Dogfood-found on `schema generate` / `sync` with `[api].auto_expose_new_tables` unset (cloud default). No tracked GitHub issue.

- [x] The linked issue is **open** and carries the `open-for-contribution` label (or I'm a Supabase maintainer).

## Checklist

- [x] Changeset added if this is a user-facing fix/feature (`bunx changeset`)

## Test plan

- [x] `cd packages/pg-delta && bun test src/plan/assumed-default-grants.test.ts src/plan/internal.test.ts src/policy/policy.test.ts`
- [x] `PGDELTA_TEST_IMAGE=postgres:17-alpine bun test tests/export-supabase-privilege-roundtrip.test.ts` (alpine dedicated cluster + gated supabase image, including identity-seq catalog pin)
- [x] `PGDELTA_TEST_IMAGE=postgres:17-alpine bun test tests/engine.test.ts` (668 corpus)
- [x] `PGDELTA_TEST_IMAGE=postgres:17-alpine bun test tests/export-fidelity.test.ts`
- [x] `bun run format-and-lint:fix && bun run check-types`


packages/pg-delta/src/plan/phases/action-emitter.ts
```diff
@@ -305,6 +347,32 @@ export function emitActions(input: ActionEmitterInput): ActionEmitterOutput {
     }
   }
 
+  // Overlay ADP wipes: export-only. The load dest can inject defaults the
+  // pristine source never had; DB-to-DB source already is the live target.
+  for (const tuple of overlayAdpWipes) {
```
avallete, line 377: **Identity-column sequences still inherit the dest injectee in the live-OFF → dest-ON case.**

I reproduced the dogfood scenario on `postgres:17-alpine`: live has no auto-expose ADP and contains `people (id bigint GENERATED ALWAYS AS IDENTITY, ...)`, `counter (n serial)` and a function; dest has `GRANT ALL ON TABLES/SEQUENCES/FUNCTIONS TO anon, authenticated, service_role`. After `load(export(live))`, tables, the function, the serial sequence and the ADP itself are all clean, but `people_id_seq` ends up with:

```
anon           SELECT,UPDATE,USAGE
authenticated  SELECT,UPDATE,USAGE
service_role   SELECT,UPDATE,USAGE
```

Two things combine: identity sequences are not facts (`relations.ts` anti-joins `deptype = 'i'`), so there is no per-object hygiene REVOKE for them (and the managed-view snapshot comparison in the new test cannot see them either), and `public/default_privileges.sql` is the *last* entry in the manifest `loadOrder`, so these wipes run after `CREATE TABLE` has already injected the grants.

Cheapest mitigation that fits this PR's mechanism: hoist the per-schema `default_privileges.sql` (or at least the wipe statements) ahead of the object files in the export order, so the wipes neutralize injection before anything is created and the per-object REVOKEs become defense in depth. If you would rather defer, please soften the changeset / PR summary and add a note under P2c in `docs/roadmap/pg-delta-next-follow-ups.md`, since the overlay makes this the one remaining leak in the exact scenario the PR fixes.

---

packages/pg-delta/src/plan/plan.ts
```diff
@@ -655,6 +665,25 @@ export function plan(
...
     : undefined;
+  const assumedDefaultGrants = uniqueAssumedDefaultGrants([
+    ...(flatPolicy?.assumedDefaultGrants ?? []),
+    ...(options?.assumedDefaultGrants ?? []),
+  ]);
+  // Export (`options.assumedDefaultGrants`) always REVOKEs overlay grantees —
+  // the load dest is an auto-expose baseline. Policy-only DB-to-DB only
+  // REVOKEs roles the apply target already has: assumedRoles skip the
+  // requirement guard, and alpine/stock Postgres has no `anon`.
+  const exportOverlay = (options?.assumedDefaultGrants ?? []).length > 0;
```
avallete, line 677: **Mode is inferred from which channel the tuples arrive through.** The same `AssumedDefaultGrant[]` means "export: REVOKE every overlay grantee and emit ADP wipes" when passed via `options`, and "hygiene only for roles present on the raw source, no ADP wipes" when it comes from the policy. `plan()` is public API and `AssumedDefaultGrant` is now re-exported from it, so a library caller passing tuples through options gets ADP wipes they probably did not ask for.

Consider making the target explicit instead of `length > 0` (e.g. an overlay option carrying the tuples plus a `target: "fresh-baseline" | "live"` discriminator, or a boolean alongside the tuples). That also removes the need for `overlayAdpWipes` and `overlayHygieneGrantees` to be derived from the same array in two different ways.

---

packages/pg-delta/src/plan/plan.ts
```diff
@@ -268,6 +274,10 @@ export interface PlanOptions {
    *  over an already-resolved view. */
   assumedSchemas?: string[];
   assumedRoles?: string[];
+  /** Overlay tuples for create-time REVOKE / ADP wipes. Distinct from
+   *  `assumedRoles` / `assumedSchemas` (those exempt the requirement guard).
+   *  Empty/absent → no overlay (projected ADP hygiene only). */
```
avallete: Nit: "Empty/absent → no overlay" is not quite right. When `options.policy` carries `assumedDefaultGrants` (as `supabasePolicy` now does), an absent option still yields overlay hygiene REVOKEs on created objects; what is absent is only the ADP wipes and the "REVOKE roles the source lacks" behavior.

---

packages/pg-delta/src/policy/supabase.ts
```diff
@@ -274,6 +274,18 @@ export const supabasePolicy: Policy = {
   // into the diff. Mirrors the system-schema exclusion list by construction.
   assumedSchemas: [...SUPABASE_SYSTEM_SCHEMAS],
 
+  // Create-time grants a fresh auto-expose-ON baseline injects. Export plans
+  // against these tuples so `load(export(live))` can REVOKE injectees live
+  // does not keep. Privilege lists stay out of the planner — only identity.
+  assumedDefaultGrants: (["r", "S", "f"] as const).flatMap((objtype) =>
```
avallete, line 280: **Every supabase-profile export now ships nine unconditional ADP REVOKEs.** A plan with source = desired = `{ schema public }` plus this overlay produces 9 `ALTER DEFAULT PRIVILEGES … REVOKE ALL … FROM …` actions; with auto-expose ON on the live side the file doubles into wipe-then-grant pairs (18 statements). This is correct and no-op safe (Postgres drops a per-schema entry that reduces to empty), but users will see these lines in their declarative schema with nothing explaining them. A one-line header comment on `default_privileges.sql`, or a docs note, would help.

Nit: the wipes are emitted one grantee per statement; the new `mergeCoTargetRevokes` only merges object-level REVOKEs, so a `REVOKE ALL ON TABLES FROM "anon", "authenticated", "service_role"` merge for ADP wipes would cut the file to a third.

---

packages/pg-delta/corpus/dependencies-cycles--partitioned-parent-publication-via-root/b.sql
```diff
@@ -0,0 +1,24 @@
+-- CLI-2299: partitioned parent, ON ONLY index, s2 partitions, via_root publication.
```
avallete, line 1: This scenario (commit df50fe4, partitioned parent + `publish_via_partition_root`) is unrelated to the privilege round-trip and is not mentioned in the PR description. Split it into its own `test(pg-delta):` PR, or at least call it out in the summary. Also, the header carries a ticket id as provenance; a behavioral description of what the scenario locks would serve the next reader better.

---

packages/pg-delta/tests/export-supabase-privilege-roundtrip.test.ts
```diff
@@ -0,0 +1,310 @@
...
+const extraClusters: Cluster[] = [];
+afterAll(async () => {
+  await Promise.all(extraPools.map((p) => p.end().catch(() => {})));
+  await Promise.all(dbs.map((d) => d.drop().catch(() => {})));
+  await Promise.all(extraClusters.map((c) => c.stop().catch(() => {})));
+});
+
+const AUTO_EXPOSE_ADP = `
+ALTER DEFAULT PRIVILEGES FOR ROLE postgres IN SCHEMA public
+  GRANT SELECT, INSERT, UPDATE, DELETE ON TABLES TO anon, authenticated, service_role;
```
avallete, line 41: The fixture understates the real injectee: the Supabase baseline is `GRANT ALL` (the committed base-init fixture shows `MAINTAIN, REFERENCES, TRIGGER, TRUNCATE` on TABLES too). The strict-subset guard compares against `_ownerDefault`, so the logic holds either way, but with DML-only here the `full_grant` table never exercises the "holder equals the owner default → elide the REVOKE" path. Suggest `GRANT ALL ON TABLES / SEQUENCES / FUNCTIONS` so the alpine cell mirrors the platform.

---

packages/pg-delta/tests/export-supabase-privilege-roundtrip.test.ts
```diff
@@ -0,0 +1,310 @@
...
+        SELECT r.rolname AS creating_role,
+               n.nspname AS schema,
+               d.defaclobjtype::text AS objtype,
+               COALESCE(g.rolname, 'PUBLIC') AS grantee
+        FROM pg_catalog.pg_default_acl d
+        JOIN pg_catalog.pg_namespace n ON n.oid = d.defaclnamespace
+        JOIN pg_catalog.pg_roles r ON r.oid = d.defaclrole
+        CROSS JOIN LATERAL aclexplode(d.defaclacl) a
+        LEFT JOIN pg_catalog.pg_roles g ON g.oid = a.grantee
+        WHERE r.rolname = 'postgres' AND n.nspname = 'public'
```
avallete, line 375: Low priority: filtering to `r.rolname = 'postgres'` means a new `public` injectee under another creating role (the image also carries `FOR ROLE supabase_admin` ADPs) would never trip this pin. Since the load runs as `postgres` this is fine today; dropping the role filter and asserting every non-owner grantee is covered would make the pin catch image drift.

---

packages/pg-delta/tests/export-supabase-privilege-roundtrip.test.ts
```diff
@@ -0,0 +1,310 @@
...
+        .filter((f) => f.id.kind === kind)
+        .map((f) =>
+          JSON.stringify({
+            id: f.id,
+            privileges: f.payload["privileges"],
+            grantable: f.payload["grantable"],
+          }),
+        )
+        .sort();
+    expect(privilegeSnap(srcView, "acl")).toEqual(
```
avallete, line 202: This comparison runs on the managed view, so it is blind to anything the fact base does not model (identity sequences, see the emitter thread). A catalog-level check alongside it (e.g. `aclexplode(relacl)` over `public` on both sides) would have caught the identity-sequence leak and keeps the test honest about "live ACLs preserved".

---

packages/pg-delta/src/plan/plan.guard.test.ts
```diff
@@ -40,13 +40,15 @@ const KIND_LITERAL_BASELINE: Readonly<Record<string, number>> = {
   // 30 → 29: the ADP gate's `desired.facts()` rescan became a prebuilt objtype
   // index (buildAdpIndex), dropping the redundant `Extract<StableId, { kind:
   // "defaultPrivilege" }>` cast — narrowing on the `!==` check suffices.
-  "internal.ts": 29,
+  // 29 → 37: overlay REVOKE elision + mergeCoTargetRevokes.
+  "internal.ts": 37,
```
avallete, line 48: Nit: +8 kind literals in one file is a big bump for this guard. Several come from the repeated `(ownerEdge.to as { kind: "role"; name: string }).name` cast pattern in `ownerDefaultPrivileges` / `revokeAllOf`; a small shared `roleNameOf(id)` helper would keep most of them out of the count.

---

avallete (commented):
Deep review of b67e17c. The design is sound: overlay tuples on the policy, planner free of platform names, strict-subset guard against `_ownerDefault`, hygiene gated on roles the apply target actually has. One gap should be closed or documented before merge; the rest are design and test-quality notes.

**Verified locally on the PR head**
- `bun test src/` (1412 pass), `check-types`, `format-and-lint`, `knip`: green.
- `tests/export-supabase-privilege-roundtrip.test.ts` on `postgres:17-alpine`: green.
- Extra hand-built plans and two Docker round-trips: column-level-only grant to `authenticated` keeps its column privilege (hygiene REVOKE orders before the column GRANT); views, serial sequences and functions round-trip cleanly; policy-only DB-to-DB on stock alpine emits no `FROM "anon"`; the fake empty-privilege ADP fact renders `REVOKE ALL` via `isRevokedDefaultMarker`; `mergeCoTargetRevokes` cannot capture an ACL REVOKE/GRANT pair because ACL attributes use replace semantics.

**Findings (details inline)**
1. Identity-column sequences still inherit the dest injectee when live is auto-expose OFF and dest is ON: no fact → no hygiene REVOKE, and `default_privileges.sql` loads last so the ADP wipes run after `CREATE TABLE`. Hoist the wipes ahead of object files, or soften the changeset and record under P2c.
2. Export vs policy-only mode is inferred from which channel the tuples arrive through; make it an explicit option on public `plan()`.
3. Every supabase-profile export now ships nine unconditional ADP REVOKEs (18 with auto-expose ON); worth a header comment or doc note, and the wipes could be merged per objtype.
4. The partitioned-parent corpus scenario (df50fe4) is unrelated to this fix.
5. Test fixture grants DML only where the platform grants ALL, so the full-set elision path is not exercised; the managed-view snapshot is blind to unmodeled identity sequences.
6. Minor: kind-literal guard bump, `PlanOptions` doc wording, `postgres`-only pin in the gated test.

---

packages/pg-delta/src/plan/phases/action-emitter.ts
avallete: Fixed in dba91c69. By-object export now gives `default_privileges.sql` a numeric sort key just before the earliest object file (pairwise ADP-vs-table swaps were not a total order with `firstAt`). Identity sequences stay unmodeled; dest auto-expose no longer injects onto them at `CREATE TABLE` time.

Coverage: file-order pin in `export-schema-adp.test.ts`; alpine catalog pin on `people_id_seq` in `export-supabase-privilege-roundtrip.test.ts`. P2c still won’t-fix per-object seq grant diffs — updated in `docs/roadmap/pg-delta-next-follow-ups.md`.

packages/pg-delta/src/plan/plan.ts
avallete: Deferred. Export vs policy-only is inferred from the channel today, and the two callers already pass the right one (schema-export sets `options.assumedDefaultGrants`; DB-to-DB uses policy). An explicit discriminator is a public-API change — recorded under PR #475 triage in `docs/roadmap/pg-delta-next-follow-ups.md`.

packages/pg-delta/src/plan/plan.ts
avallete: Fixed in dba91c69. `PlanOptions.assumedDefaultGrants` JSDoc now says an absent option still applies policy hygiene; ADP wipes and “REVOKE roles the source lacks” are options-only.

packages/pg-delta/src/policy/supabase.ts
avallete: Deferred. The wipes are correct and no-op safe. A header comment or per-objtype merge is presentation — recorded under PR #475 triage in `docs/roadmap/pg-delta-next-follow-ups.md`.

packages/pg-delta/corpus/dependencies-cycles--partitioned-parent-publication-via-root/b.sql
avallete: Called out in the PR summary. Leaving df50fe4 on this branch rather than splitting; the scenario header already describes partitioned parent + `publish_via_partition_root`.

packages/pg-delta/tests/export-supabase-privilege-roundtrip.test.ts
avallete: Deferred. The alpine live SQL is DML; full-set elision is covered in `assumed-default-grants.test.ts`. Recorded under PR #475 triage in `docs/roadmap/pg-delta-next-follow-ups.md`.

packages/pg-delta/tests/export-supabase-privilege-roundtrip.test.ts
avallete: Deferred as low priority. Load still runs as `postgres`; recorded under PR #475 triage in `docs/roadmap/pg-delta-next-follow-ups.md`.

packages/pg-delta/tests/export-supabase-privilege-roundtrip.test.ts
avallete: The new alpine identity-seq test queries `aclexplode(relacl)` on `people_id_seq` (the unmodeled leak). Not adding a full-public catalog dump on the DML fixture — recorded under PR #475 triage in `docs/roadmap/pg-delta-next-follow-ups.md`.

packages/pg-delta/src/plan/plan.guard.test.ts
avallete: Deferred. The kind-literal bump is from overlay wipe/hygiene guards; a `roleNameOf` helper is cleanup, not a defect. Recorded under PR #475 triage in `docs/roadmap/pg-delta-next-follow-ups.md`.

packages/pg-delta/src/plan/internal.ts
```diff
@@ -1291,6 +1330,15 @@ export function elideCoCreateRevokeBeforeGrant(
...
     if (defaultGrantsOutside(aclId.target, aclId.grantee, new Set(privileges)))
       return; // REVOKE is load-bearing
+    const matches = overlayMatches(
+      assumedDefaultGrants,
+      aclId.target,
+      aclId.grantee,
+    );
+    if (matches.length > 0) {
+      const fullSet = ownerDefaultPrivileges(desired, aclId.target);
+      if (fullSet === undefined || isStrictSubset(privileges, fullSet)) return; // dest injectee may be a superset of desired ADP / object ACL
```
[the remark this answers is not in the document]
avallete: Deferred. Overlay tuples do not carry grant-option bits, so when the privilege names equal `_ownerDefault` we still drop the leading `REVOKE`. A dest ADP that injected the same privileges `WITH GRANT OPTION` would survive a later plain `GRANT`. That is not the Supabase auto-expose injectee, and we already keep the REVOKE when `payload.grantable` is non-empty on the desired side. Recorded in docs/roadmap/pg-delta-next-follow-ups.md § PR #475 rather than modeling grant options on overlay tuples in this PR.

packages/pg-delta/src/frontends/export-sql-files.ts
```diff
@@ -1048,9 +1056,32 @@ export function exportSqlFiles(
     files.set(path, entry);
   });
 
-  const ordered = [...files.entries()].sort(
-    (a, b) => a[1].firstAt - b[1].firstAt,
+  // Overlay ADP wipes must land before CREATE: dest auto-expose injects
+  // grants onto unmodeled identity sequences, and ALTER DEFAULT PRIVILEGES
+  // does not revoke existing object ACLs. Numeric keys (not ADP-vs-object
+  // pairwise swaps): a swap comparator is not transitive with firstAt.
```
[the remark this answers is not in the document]
avallete: Deferred. `--layout ordered` still emits files in raw plan order (the hoist lives in the by-object path only). That is already recorded in the changeset and in docs/roadmap/pg-delta-next-follow-ups.md § PR #475. Identity-sequence inject is a by-object / schema-first concern; we are not expanding this PR to hoist ADP in ordered layout.

jgoux (approved):

packages/pg-delta/src/frontends/export-sql-files.ts
```diff
@@ -1048,9 +1056,32 @@ export function exportSqlFiles(
...
+    ...[...files.entries()]
+      .filter(([path]) => isObjectFile(path))
+      .map(([, entry]) => entry.firstAt),
+    Number.POSITIVE_INFINITY,
   );
+  const ordered = [...files.entries()].sort((a, b) => {
+    const key = (path: string, firstAt: number): number =>
+      path.endsWith("/default_privileges.sql")
+        ? Math.min(firstAt, objectMinFirstAt) * 2 - 1
+        : firstAt * 2;
```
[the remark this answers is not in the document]
avallete: Confirmed on `postgres:17-alpine` at f4e0278, raw profile, no overlay involved. Source:

```sql
CREATE SCHEMA app;
CREATE TABLE app.t1 (id integer);
CREATE TABLE app.ident (id bigint GENERATED ALWAYS AS IDENTITY);
ALTER DEFAULT PRIVILEGES IN SCHEMA app GRANT SELECT ON TABLES TO rvc_reader;
ALTER DEFAULT PRIVILEGES IN SCHEMA app GRANT USAGE ON SEQUENCES TO rvc_reader;
CREATE TABLE app.t2 (id integer);
CREATE TABLE app.ident2 (id bigint GENERATED ALWAYS AS IDENTITY);
```

Export → load, then `aclexplode(relacl)` for `rvc_reader` in `app`:

```
SRC : ident2 SELECT, ident2_id_seq USAGE, t2 SELECT
DEST: ident2 SELECT, ident2_id_seq USAGE, ident_id_seq USAGE, t2 SELECT
```

`t1` and `t2` are correct now that hygiene uses the default owner. `ident_id_seq` is the leak this thread describes: the positive `GRANT USAGE ON SEQUENCES` is hoisted with the wipes, so the predating identity sequence is created under it and nothing modeled removes the grant. On `main` this shape was correct and the opposite one (`ident2`) leaked, so the whole-file hoist swaps which history diverges rather than fixing either.

Suggestion that closes this and the extension thread together: hoist only the REVOKE-only ADP statements (the overlay wipes) into their own file that sorts before object files and `_cluster/extensions/*`, and leave positive ADP GRANTs at their plan position. Wipes depend only on roles and the schema, so nothing else has to move.

---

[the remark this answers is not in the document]
avallete: Duplicate of the thread above; reproduction and suggested fix are there: https://github.com/supabase/pg-toolbelt/pull/475#discussion_r4004472101

---

packages/pg-delta/src/frontends/export-sql-files.ts
```diff
@@ -1048,9 +1062,32 @@ export function exportSqlFiles(
     files.set(path, entry);
   });
 
-  const ordered = [...files.entries()].sort(
-    (a, b) => a[1].firstAt - b[1].firstAt,
+  // Overlay ADP wipes must land before CREATE: dest auto-expose injects
+  // grants onto unmodeled identity sequences, and ALTER DEFAULT PRIVILEGES
+  // does not revoke existing object ACLs. Numeric keys (not ADP-vs-object
+  // pairwise swaps): a swap comparator is not transitive with firstAt.
+  const objectDirs = new Set([...Object.values(SCHEMA_DIRS), "indexes"]);
```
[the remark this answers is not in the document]
avallete: Confirmed on `postgres:17-alpine` at f4e0278 with the supabase profile. Source (no ADP): `CREATE EXTENSION pg_trgm SCHEMA public; CREATE TABLE public.t (id int);`. Destination: the auto-expose `GRANT ALL ON TABLES/SEQUENCES/FUNCTIONS TO anon, authenticated, service_role` in `public`.

Export load order:

```
_cluster/extensions/pg_trgm.sql
public/default_privileges.sql
public/tables/t.sql
public/schema.sql
```

Count of `anon` grants on `public` functions named `%trgm%` after load:

```
src 0, dest 16
```

The extension is installed before the wipe runs, its members inherit the destination default grants, and since member ACLs are recorded only as deviations from `pg_init_privs` no satellite exists to remove them. Exposure is limited to extensions installed into an overlay schema, but that is an ordinary `CREATE EXTENSION`.

Same fix as the identity-sequence thread: give the REVOKE-only wipes a file that sorts before `_cluster/extensions/*` as well as object files, while positive ADP GRANTs keep their plan position.

---

packages/pg-delta/src/frontends/export-sql-files.ts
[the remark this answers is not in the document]
avallete: Fixed in b9426481. Overlay `REVOKE ALL` is now `<schema>/adp_wipes.sql`, sorted after `schema.sql` / roles and before object files and `_cluster/extensions/*`. Positive GRANT ADP stays in `default_privileges.sql` at plan `firstAt`.

Alpine pin (raw profile): `ident_id_seq` has no `rvc_reader` USAGE; `t2` still has SELECT. `ident2_id_seq` USAGE is unmodeled (create-order is not in the fact base) — recorded under P2c in docs/roadmap/pg-delta-next-follow-ups.md § PR #475.

packages/pg-delta/src/frontends/export-sql-files.ts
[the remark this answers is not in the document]
avallete: Fixed in b9426481. Wipe files sort before `_cluster/extensions/*` as well as object files. Alpine + supabase profile: `CREATE EXTENSION pg_trgm SCHEMA public` then a table, dest auto-expose ON — 0 `anon` grants on `%trgm%` functions after load.

packages/pg-delta/src/frontends/export-sql-files.ts
```diff
@@ -1048,9 +1062,32 @@ export function exportSqlFiles(
...
+  const objectMinFirstAt = Math.min(
+    ...[...files.entries()]
+      .filter(([path]) => isObjectFile(path))
+      .map(([, entry]) => entry.firstAt),
+    Number.POSITIVE_INFINITY,
   );
+  const ordered = [...files.entries()].sort((a, b) => {
+    const key = (path: string, firstAt: number): number =>
+      path.endsWith("/default_privileges.sql")
+        ? Math.min(firstAt, objectMinFirstAt) * 2 - 1
```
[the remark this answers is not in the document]
avallete: Fixed in b9426481. Global `REVOKE ALL` is `_cluster/adp_wipes.sql`; matching GRANT ADP is `_cluster/default_privileges.sql` (not mixed into `roles.sql`), with the wipe file before the GRANT file and before objects.

packages/pg-delta/src/frontends/export-sql-files.ts
```diff
@@ -988,7 +1016,15 @@ export function exportSqlFiles(
   const miscPath = `${clusterRoot(pathStyle)}/misc.sql`;
   const actionPaths = rendered.actions.map((action) => {
     const subject = subjectOf(action, fb);
-    return subject === undefined ? miscPath : pathFor(subject, pathContext);
+    if (subject === undefined) return miscPath;
+    if (
+      layout !== "ordered" &&
```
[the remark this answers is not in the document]
avallete: Deferred. `--layout grouped` still category-orders all ADP as cluster, so schema-scoped wipes can run before `CREATE SCHEMA` and retry can let tables/extensions commit first. Default by-object (schema-first) is the path this PR fixes. Recorded under PR #475 triage in docs/roadmap/pg-delta-next-follow-ups.md — we are not expanding the grouped assembler here.

packages/pg-delta/src/frontends/export-sql-files.ts
```diff
@@ -988,7 +1016,15 @@ export function exportSqlFiles(
...
+      subject.kind === "defaultPrivilege" &&
+      isAdpWipeSql(action.sql)
```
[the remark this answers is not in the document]
avallete: Fixed in the follow-up commit: wipe-file routing is gated on `assumedDefaultGrants` matching the ADP id, not SQL `REVOKE ALL` alone. A desired empty ADP (`REVOKE ALL ON FUNCTIONS FROM PUBLIC`) stays in `default_privileges.sql` at plan order, after `CREATE EXTENSION`. Overlay wipes still go to `adp_wipes.sql` before objects/extensions. Unit pin in `export-schema-adp.test.ts`.

PR #480, opened 2026-09-17
fix(pg-delta): keep attached child indexes on their partition's drop root

avallete:
## Summary

Dropping a partitioned table together with its partitions cycled in `topoSort` (guardrail 4) whenever the parent carried an index. Each partition holds an attached child index; the index rule's `dropRootRedirect` folded that child into the parent index's drop root, which is the parent table's `DROP TABLE`. The child-teardown rule then ordered the parent's DROP before the partition's DROP, while the inherit edge ordered it after.

The redirect now returns nothing when the child index's own table is removed or replaced, so the child folds into its partition's `DROP TABLE`, which PostgreSQL cascades anyway. Tables carrying only a primary key were unaffected, since constraint-backed indexes are not extracted as index facts.

The regression was introduced by #470: the new scenario passes on the release before it and fails on `main`.

## Changes

- `packages/pg-delta/src/plan/rules/indexes.ts`: skip the parent-index redirect when the index's owning table is itself removed.
- `packages/pg-delta/src/plan/partition-index-attach.test.ts`: fact-base regression asserting the partition DROP followed by the parent DROP, no cycle.
- `packages/pg-delta/corpus/partitioned-table-operations--drop-family-with-indexes/`: corpus scenario dropping a whole indexed partition family (three partitions including DEFAULT, two parent indexes).
- `.changeset/partition-family-drop-cycle.md`: patch changeset.

RED output on `main` (first commit, tests only):

```
dependency cycle among 2 actions — this is a rule/emission bug, fix the rule (guardrail 4):
  DROP TABLE "s1"."p1"
  DROP TABLE "s1"."parent"
```

## Linked issue

No tracked issue; found via production error triage of the branch diff endpoint.

- [x] The linked issue is **open** and carries the `open-for-contribution` label (or I'm a Supabase maintainer).

## Checklist

- [x] Changeset added if this is a user-facing fix/feature

https://claude.ai/code/session_01EUM6tERVxNNYVhDSVgsQNh


jgoux (commented):
Reviewed all five changed files at a657d98bee2f0929524af9508912f8f629370b39 and cross-checked with an independent Claude Fable 5.1 read-only review. No actionable defects found, so there are no inline defect comments.

Validation:
- Nine attached-index planner tests passed.
- Thirty focused partition integration checks passed on PostgreSQL 17: ten corpus scenarios in both directions, plus fixture validity, covering indexed family teardown, parent-index replacement/rename, repartitioning, and multilevel partitions.
- An isolated runtime check reproduced the two-action dependency cycle with the baseline redirect rule and confirmed the PR rule produces partition DROP followed by parent DROP.
- Executed the additional case suggested by Fable (replace a partition's bound while removing its parent index): it emits DROP INDEX on the parent, DROP TABLE on the partition, and partition recreation, with no illegal attached-child index DROP.

Fable's sole finding was an optional coverage suggestion for that last case, explicitly not a current defect. I did not promote the speculative future regression risk to an inline finding. Validation was focused on PostgreSQL 17; the full suite and other PostgreSQL versions were not run.

jgoux (approved):

avallete:
Thanks for the review. The optional case you validated (replace a partition's bound while removing the parent index) is now pinned as a unit test in `f6bf207`: it asserts the parent `DROP INDEX`, the partition `DROP TABLE` and recreate, and that no statement touches the attached child index.

---

Issue #483, opened 2026-09-22
I'm trying out the new pg-delta, it seems to run considerably faster and fix some few issues previously. However, whenever I do `plan` or `diff`, I get lots of `WARNING [dangling_edge]` messages like so:


irisjae:
I'm trying out the new pg-delta, it seems to run considerably faster and fix some few issues previously. However, whenever I do `plan` or `diff`, I get lots of `WARNING [dangling_edge]` messages like so:

```
[source] WARNING [dangling_edge] constraint:public.account.account_name_not_null: edge constraint:public.account.account_name_not_null -[depends]-> column:public.account.name references a fact not in the base
[source] WARNING [dangling_edge] constraint:public.account.account_email_not_null: edge constraint:public.account.account_email_not_null -[depends]-> column:public.account.email references a fact not in the base
...
[source] WARNING [dangling_edge] constraint:schema2.co_relations.co_relations_many_collection_not_null: edge constraint:schema2.co_relations.co_relations_many_collection_not_null -[depends]-> column:schema2.co_relations.many_collection references a fact not in the base
[source] WARNING [dangling_edge] constraint:schema2.co_relations.co_relations_many_field_not_null: edge constraint:schema2.co_relations.co_relations_many_field_not_null -[depends]-> column:schema2.co_relations.many_field references a fact not in the base
...
[source] WARNING [unmodeled_kind] 1 unmodeled "statistics object" object not managed by this engine (e.g. co_event_expr_expr_expr_stat) — v1 detects but does not model this kind
[desired] WARNING [dangling_edge] constraint:schema2.co_activity.co_activity_id_not_null: edge constraint:schema2.co_activity.co_activity_id_not_null -[depends]-> column:schema2.co_activity.id references a fact not in the base
[desired] WARNING [dangling_edge] constraint:schema2.co_activity.co_activity_item_not_null: edge constraint:schema2.co_activity.co_activity_item_not_null -[depends]-> column:schema2.co_activity.item references a fact not in the base
...
[desired] WARNING [dangling_edge] constraint:public.order.order_id_not_null: edge constraint:public.order.order_id_not_null -[depends]-> column:public.order.id references a fact not in the base
[desired] WARNING [dangling_edge] constraint:public.order.order_code_not_null: edge constraint:public.order.order_code_not_null -[depends]-> column:public.order.code references a fact not in the base
```

The migration itself appears to run fine. Should I be concerned?

[...]

PR #485, opened 2026-09-22
fix(pg-delta): stop warning dangling_edge for every NOT NULL column on PostgreSQL 18

avallete:
## Summary

On PostgreSQL 18, every `pgdelta plan` / `diff` / `snapshot` emitted one warning per NOT NULL column:

```
WARNING [dangling_edge] constraint:public.account.account_name_not_null: edge constraint:public.account.account_name_not_null -[depends]-> column:public.account.name references a fact not in the base
```

Large schemas get hundreds of these, burying the diagnostics that matter (the reporter's output also contained a genuine `unmodeled_kind` warning for an extended-statistics object, lost in the noise).

**Cause.** PostgreSQL catalogs NOT NULL as a `pg_constraint` row (`contype = 'n'`) — for domains since PG 17, for **table columns since PG 18**. pg-delta models NOT NULL as an attribute on the payload (`pg_attribute.attnotnull` on the column fact), and `src/extract/relations.ts` extracts only `contype IN ('p','u','f','c','x')`. But the pg_depend resolver's `tcon` branch took every row with `conrelid <> 0`, so each NOT NULL row resolved to a `constraint:` id whose fact does not exist. `FactBase` discarded the edge and warned.

**Fix.** Exclude those rows from `tcon`, sharing one predicate (`NOT_NULL_IS_NOT_A_FACT`) with the domain-side `dcon` exclusion so the two branches cannot drift — same root cause, different PostgreSQL release.

**This is diagnostics-only.** The edge was already dropped before it reached the fact base, so the fact base, its content hashes, and every resulting plan are unchanged. `tests/depend-edges-oracle.test.ts` pins the surviving edge set with an inline snapshot across PG 14–18 and is untouched by this change.

## Stacking

> [!IMPORTANT]
> Based on #484, **not** `main`. #484 fixes the domain half of this same root cause and its `dcon` filter is the line this PR generalizes. Merge #484 first; this retargets to `main` cleanly afterwards.
>
> `tests.yml` only triggers on PRs into `main`, so **this PR gets no test CI while stacked** — see the local gate below.

## RED → GREEN

`tests/diagnostic-noise.test.ts`, on `postgres:18-alpine`, before the fix:

```text
(fail) extraction diagnostic noise (P1) > no dangling_edge diagnostics for catalog NOT NULL rows (issue #483)
(fail) extraction diagnostic noise (P1) > extraction of an ordinary user schema emits no diagnostics at all
 3 pass
 2 fail

edge constraint:app.nn.nn_id_not_null -[depends]-> column:app.nn.id references a fact not in the base
edge constraint:app.nn.nn_label_not_null -[depends]-> column:app.nn.label references a fact not in the base
edge constraint:app.nn.nn_note_not_null -[depends]-> column:app.nn.note references a fact not in the base
```

Green on PG 18 after the fix; green on PG 17 and PG 14 both before and after (neither has `contype 'n'` rows for tables).

## Tests

- `tests/diagnostic-noise.test.ts`: the fixture now carries NOT NULL in all three shapes PostgreSQL records it (implied by a PRIMARY KEY, declared inline, added via `ALTER TABLE`). Two new assertions — one matching the dangling-edge **shape** rather than the `_not_null` naming convention, one asserting a clean user schema extracts with zero diagnostics at all.
- No corpus scenario: the plan output is byte-identical before and after, so a corpus case would assert nothing this doesn't.

**Local gate** (CI can't run it while stacked):

| Check | Result |
|---|---|
| `diagnostic-noise` + `depend-edges-oracle` + `domain-not-null-constraint` on PG 18 / 17 / 14 | 12 pass each |
| `bun test src/` | 1498 pass |
| Full corpus, `postgres:18-alpine` | see comment below — still running at time of open |
| `format-and-lint` / `check-types` | clean |

## Reviewer notes

A same-harness review probed whether the diagnostics-only claim actually holds — specifically whether a `contype='n'` row could ever resolve to a fact that *does* exist. It cannot: only `relations.ts` and `types.ts` mint `constraint:` ids, a same-relation name collision is blocked by pg_constraint's unique `(conrelid, contypid, conname)`, and a domain/table name collision is impossible because the table's row type already occupies that `pg_type` name. Every pg_depend row touching a `contype='n'` constraint has it as the *dependent*, never the referenced endpoint, and `conindid = 0` so the constraint-backed-index branch is unaffected.

**Deferred, recorded in `docs/roadmap/pg-delta-next-follow-ups.md`:** `conislocal = false` rows (inherited / partition-child constraints) dangle the same way on *every* version — pre-existing, not touched here. The mechanical fix is the same shape, but whether such a constraint should be skipped or resolved to its **parent** constraint is a modeling decision rather than a one-line filter.

## Linked issue

Closes #483

- [x] The linked issue is open (Supabase maintainer)

## Checklist

- [x] Changeset added (`.changeset/table-not-null-dangling-edge.md`, patch)


avallete:
Full corpus gate, promised in the description — **green**:

```text
PGDELTA_TEST_IMAGE=postgres:18-alpine bun test tests/engine.test.ts

corpus postgres:18-alpine [686/686 100%] PASS view-operations--simple-create (b->a)

 686 pass
 0 fail
Ran 686 tests across 1 file. [745.43s]
```

Every scenario, both directions, on the affected version. Together with the untouched `depend-edges-oracle` snapshot that's the evidence for the diagnostics-only claim: no plan changed, nothing stopped converging.

Worth restating since CI can't run here — `tests.yml` is `branches: [main]`, so this stacked PR shows no test checks. The matrix runs once #484 merges and this retargets.

